PR #38, opened 2014-11-13
add endpoint list

tszpinda:
Hi,
Any chance to merge this one? 

use case:
- allows to list endpoinds required by front end additionally including example data

### DIFF
--- a/pretender.js
+++ b/pretender.js
@@ -20,6 +20,7 @@ function Pretender(maps){
   this.handlers = [];
   this.handledRequests = [];
   this.unhandledRequests = [];
+  this.endpointList = [];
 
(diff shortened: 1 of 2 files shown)


trek:
I'm not sure I understand the value in this. Can you describe more?

tszpinda:
We are building an app without back end at the moment, so we mock all the apis using pretender. When time comes to write the server side, we could easily create a page which would list all the apis mocked by pretender, so we can easily see what apis are required and what is the expected format.

stefanpenner:
the relevant code has changed dramatically. We may consider something that takes the new changes into account.  In-order to accept I believe we would require a good example that demonstrates the value.

PR #48, opened 2014-12-27
Async Responses

DingoEatingFuzz:
PR for issue https://github.com/trek/pretender/issues/41

- Makes responses asynchronous by default
- Responses resolve on the next run of the event loop by default
- Responses can be made synchronous by passing `false` for the async argument when specifying a route
```js
pretender.get('/sync/thing', handlerFn, false)
```
- Responses can be set to resolve after a specified time by passing in a number for the async argument
```js
pretender.get('/timed/thing', handlerFn, 300) // in ms just like setTimeout
```
- Responses can be set to never automatically resolve by passing in `true` for the async argument
```js
pretender.get('/manual/thing', handlerFn, true);
// ...
pretender.resolve(request);
```
- Responses can be set to conditionally resolve by passing in a function for the async argument
```js
pretender.get('/maybe/async/thing', handlerFn, function() { return isAsync ? 0 : false; })
```
- A helper by the name of `pretender.requiresManualResolution` exists to check if a request will, well, require manual resolution.

### DIFF
--- a/pretender.js
+++ b/pretender.js
@@ -21,6 +21,7 @@ function Pretender(maps){
   this.handledRequests = [];
   this.passthroughRequests = [];
   this.unhandledRequests = [];
+  this.requestReferences = [];
 
   // reference the native XMLHttpRequest object so
   // it can be restored later
@@ -112,8 +113,8 @@ function interceptor(pretender) {
 }
 
 function verbify(verb){
-  return function(path, handler){
-    this.register(verb, path, handler);
+  return function(path, handler, async){
+    this.register(verb, path, handler, async);
   };
 }
 
@@ -140,12 +141,13 @@ Pretender.prototype = {
   'delete': verbify('DELETE'),
   patch: verbify('PATCH'),
   head: verbify('HEAD'),
-  register: function register(verb, path, handler){
+  register: function register(verb, path, handler, async){
     if (!handler) {
       throw new Error("The function you tried passing to Pretender to handle " + verb + " " + path + " is undefined or missing.");
     }
 
     handler.numberOfCalls = 0;
+    handler.async = async;
     this.handlers.push(handler);
 
     var registry = this.registry[verb];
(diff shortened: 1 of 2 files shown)


pretender.js
```diff
@@ -178,24 +179,66 @@ Pretender.prototype = {
...
+      });
+
+      if (strategy !== true) {
+        setTimeout(function() {
+          pretender.resolve(request);
+        }, typeof strategy === 'number' ? strategy : 0);
+      }
+    }
+  },
+  resolve: function resolve(request) {
```
trek: What's `fn`?

DingoEatingFuzz: `fn` is an ambiguous variable name :stuck_out_tongue: 

It's the callback that handles the response. I will rename it to something better.

DingoEatingFuzz:
@trek, this is ready for review again (not sure if new commits emit notifications).

pretender.js
```diff
@@ -21,6 +21,7 @@ function Pretender(maps){
   this.handledRequests = [];
   this.passthroughRequests = [];
   this.unhandledRequests = [];
+  this.requestRefs = [];
```
trek: This existing code doesn't tend to use abbreviations. This ref says "Yellow Card"

pretender.js
```diff
@@ -178,24 +180,65 @@ Pretender.prototype = {
...
+            pretender = this;
 
-        this.handledRequest(verb, path, request);
+        this.handleResponse(request, async, function() {
+          request.respond(status, headers, body);
+          pretender.handledRequest(verb, path, request);
+        });
       } catch (error) {
         this.erroredRequest(verb, path, request, error);
+        pretender.resolve(request);
```
trek: I'd prefer to keep this as `this`.

trek:
Looks good other than a few style choices. I think once those are fixed, squash into a single feature commit for merge.

pretender.js
DingoEatingFuzz: Change it to `requestReferees`, got it :trollface:

DingoEatingFuzz:
All squashed up.

trek:
Can you update the README? The docs are probably all wrong now.

DingoEatingFuzz:
Yep, working on those changes now.

PR #51, opened 2015-01-15
Update README to explain the async features

DingoEatingFuzz:
Please proofread :grin:


trek:
Some small tweaks I'd do, but I we'll fix in post.

PR #58, opened 2015-01-30
Add jspm support

guybedford:
This shim is necessary to work for jspm because otherwise it runs as a CommonJS module, while failing the `isNode` check, resulting in the dependencies not being picked up.

An alternative approach would be to change the `isNode` function to accept the browserify process shim, perhaps by checking a standard property.

No pressure to take this either, we have an override that's making it work otherwise.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,5 +29,13 @@
   "dependencies": {
     "fake-xml-http-request": "^1.2.0",
     "route-recognizer": "^0.1.9"
+  },
+  "jspm": {
+    "shim": {
+      "pretender": {
+        "deps": ["route-recognizer", "fake-xml-http-request"],
+        "exports": "Pretender"
+      }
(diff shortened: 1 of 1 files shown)


stefanpenner:
@guybedford will gladly merge this if rebased.

guybedford:
@stefanpenner no problem, have done. CI seems not to be happy though unfortunately, but don't think it's related. Thanks for the follow-up sir!

stefanpenner:
i believe we are green: https://github.com/pretenderjs/pretender/pull/100

stefanpenner:
@guybedford FYI, we plan to cut a release shortly. Would love to sneak this in.

guybedford:
Sure, should be there now.

stefanpenner:
woot :)

PR #62, opened 2015-04-13
Remove setting of timeout on synchronous request obj

meismann:
This line caused an error like "Failed to set the timeout
property on XMLHttpRequest: Timeouts cannot be set for
synchronous requests made from a document."

I get this message in uptodate Chrome 41.0.2272.118 (64-bit)
and Firefox 37.0.1. My German FF prints the message in
German, so this must be a browser-side restriction.


trek:
Although it fixes your issue, I believe this will cause other issues. Can you also submit tests around this?

stefanpenner:
this seems to have gone stale, will gladly re-consider if tests are provided

PR #73, opened 2015-05-06
Simulate progress events

cibernox:
While a request is not resolved or aborted, `onprogress` events will be fired each 50ms.

@trek I don't think that there is any risk of leaking. As soon the request is not active not more progress will be fired.

### DIFF
--- a/pretender.js
+++ b/pretender.js
@@ -135,37 +135,15 @@ function throwIfURLDetected(url){
   }
 }
 
-/*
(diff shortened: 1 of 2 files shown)


knownasilya:
Ping

cibernox:
This is finished on my side unless anyone spots anything funky

cibernox:
Wooot!

PR #74, opened 2015-05-21
[help wanted] Fix passthrough bug in phantomjs

samselikoff:
Pretender's test suite fails in phantomjs:

> InvalidStateError: DOM Exception 11: An attempt was made to use an object that is not, or is no longer, usable.

It's the `passthrough` test. I haven't been able to get it to pass.


samselikoff:
I found this SO thread which talks about `async: false`: http://stackoverflow.com/a/5319647/1406664. I tried setting async false in the passthrough code, to no avail.

Any ideas?

samselikoff:
closing in favor of #76

PR #66, opened 2015-04-15
[Ehancement] Host support

mike-north:
Route-recognizer currently only supports same-origin requests. @stefanpenner and I paired on this to add support for CORS api mocking in our tests

### TODO
- [x] Tests!

PR #75, opened 2015-06-05
Adding passthroughURL to allow all requests for URL to be passed through

step2yeung:
In our tests, we need to send an URL request out to a express server. With pretender as it is currently, we do not have a way to to do this.
This flag would allow URL requests to be passed through.

### DIFF
--- a/pretender.js
+++ b/pretender.js
@@ -35,6 +35,9 @@ function Pretender(/* routeMap1, routeMap2, ...*/){
   // "start" the server
   this.running = true;
 
+  // allows for URL requests to passthrough
+  this.passthroughURL = false;
+
   // trigger the route map DSL.
   for(i=0; i < arguments.length; i++){
     this.map(arguments[i]);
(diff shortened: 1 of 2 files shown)


trek:
Can you explain a bit how this is different than the existing passthrough behaviors?

step2yeung:
Sure, I added this to deal with fully qualified URLs. Since pretender only handle requests to paths, in checkPassthrough, my requests will fail at 'throwIfURLDetected' since it is a fully qualified URLs.

I think in the spirit of pretender, this change will still respect the basis of only handling requests to path, while allowing anything with fully qualified URLs to go thru if required.

trek:
Even if https://github.com/trek/pretender/pull/66 is merged?

step2yeung:
Thanks Trek, that PR is exactly that's needed.

trek:
Ok, closing in favor of #66 then

PR #76, opened 2015-06-09
Add phantom to karma and fix passthrough test timing issue.

cball:
There were 2 things going on:
* phantomjs and `onreadystatechange` do not like each other. Simply trying to access `xhr['onreadystatechange']` would throw an error. This PR uses `onload` if its available with a fallback to `onreadystatechange` if it's not.
* the `setTimeout` call w/ a 0 delay was coming back after the ajax request had already completed in phantom. I bumped the ajax request response timing up to 100 ms to avoid issues.

Thanks to @kategengler for pointing out a [similar issue in mockjax](https://github.com/jakerella/jquery-mockjax/blob/master/src/jquery.mockjax.js#L222) which helped narrow down the root problem.

Fixes #74

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -60,7 +60,7 @@ module.exports = function(config) {
 
     // start these browsers
     // available browser launchers: https://npmjs.org/browse/keyword/karma-launcher
-    browsers: ['Chrome'],
+    browsers: ['Chrome', 'PhantomJS'],
 
 
     // Continuous Integration mode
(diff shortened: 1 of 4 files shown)


cball:
@trek do you have any comments on this?

samselikoff:
Let me know if I can help here. I'd love to see this merged as it's a dependency for a mirage feature.

trek:
Thanks!

samselikoff:
<3 awesome, thx so much! Hope you had a relaxing time :)

brzpegasus:
@trek Are you planning on publishing a new release anytime soon? We're hitting the same passthrough timing issue and would love to get this fix in (other than pulling from master).

PR #78, opened 2015-06-23
Fix wrong invocation of dispatchEvent method for passthrough requests.

simonihmig:
This fixes a bug where event listeners added via addEventListener() to a passthrough request were not being called.

### DIFF
--- a/pretender.js
+++ b/pretender.js
@@ -81,7 +81,7 @@ function interceptor(pretender) {
           }
         }
         // fire fake events where applicable
-        fakeXHR.dispatchEvent(evt, e);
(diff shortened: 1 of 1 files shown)


stefanpenner:
this will need a test

samselikoff:
@simonihmig this still relevant?

simonihmig:
I think so, the signature for dispatchEvent expects a single argument, the event object, but it is called as `fakeXHR.dispatchEvent(evt, e);`

I didn't have the time to write a test for getting this merged though, I will try to do it the next days...

bantic:
Thanks for pointing this out. This was fixed in #117.

PR #79, opened 2015-06-23
Support responseType for passthrough requests

simonihmig:
Currently the responseType of the fake xhr is ignored, so any special responseType like "arraybuffer" does not work. This will fix it. Probably also https://github.com/trek/pretender/issues/72 if passthrough requests are involved.


stefanpenner:
@simonihmig if we are to accept it it will need thorough test coverage.

AlexandreBonaventure:
Is this going to be merged in master ?
Thanks

Issue #52, opened 2015-01-19
ES6


trek:
Move Pretender (and all its dependencies) to ES6 style modules. 

- [x] https://github.com/tildeio/route-recognizer/
- [ ] https://github.com/trek/FakeXMLHttpRequest

Tests:
- [ ] how do you even run tests like this? transpile pre-run?

Publish:
- [ ] CJS to node
- [ ] ES6 to bower
- [ ] AMD, sorry, you'll need to compile yourself?

PR #86, opened 2015-07-25
Convert to ES6

code0100fun:
**Addresses [#52](https://github.com/trek/pretender/issues/52)**

This is a first pass at upgrading to ES6. I also converted from karma to testem :)

I left the single file `pretender.js` file intact, not sure how you want to split it up.

**Concerns:**
* Single UMD module may not be the best approach
* Tests are not ran against node.js
* Versioning `dist/pretender.js` may not be desired

**Todo:**
* Could use ember-cli to run/build


mike-north:
> Could use ember-cli to run/build

There's a bug in testem where it doesn't respect the `cwd` (working directory) passed to it. Once I can get that fixed, it'll be much simpler than what you have here to use ember-cli for basic JavaScript libraries.

Also I'm not sure how well `broccoli-cli` works on windows, whereas I know ember-cli works (pretty much) fine

trek:
Thanks for this, but I think we're holding off. The landscape just keeps shifting

PR #90, opened 2015-08-04
cleanup

stefanpenner:
* use `self` so it is more likely to work in a SW/WW
* ‘use strict’ to prevent previous leaks
* fix use strict violations
* add missing semi-colons
* mixup whitespace
* remove unused slice


stefanpenner:
depends on: https://github.com/trek/pretender/pull/89

PR #96, opened 2015-08-06
Remove node

stefanpenner:
this depends on: https://github.com/pretenderjs/pretender/pull/95

relevant commit: https://github.com/pretenderjs/pretender/commit/71dbac0a240fc202fb90b95d5b8516eb2f55f2eb

Why?:

* the current code does not support node, it uses window/self all over the place
* the current node detection breaks in NW.js environments
* the code isn't tested on node
* node doesn't have `global.XMLHTTPRequest` 
  * in theory, we attempt to patch https://github.com/driverdan/node-XMLHttpRequest, but do to how NPM dedupes stuff, this is likely fraught with peril.

@trek this may be controversial so I would like your +1

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,13 +1,17 @@
 {
   "predef": [
-    "self"
+    "self",
+    "require",
+    "module",
+    "define",
+    "process"
   ],
   "expr": true,
   "proto": true,
   "strict": true,
   "indent": 2,
   "camelcase": true,
-  "node": true,
+  "node": false,
   "browser": true,
   "boss": true,
   "curly": true,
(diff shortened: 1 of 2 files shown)


mike-north:
rebase please

stefanpenner:
rebased

stefanpenner:
@trek wants me to ensure browserify works correctly, then we can :ship:.

mike-north:
LGTM

trek:
I'm +1

taras:
@stefanpenner @trek would you guys be open to a proper implementation of node support? #192

PR #98, opened 2015-08-06
add travis

stefanpenner:


mike-north:
LGTM. Will add phantomjs2 as part of a build matrix in a follow up

Issue #70, opened 2015-04-29
Middleware or filters


saarangsoltani:
it would be very handy to have some sort of filtering functionality where a function would be executed before a request handler and if that function returned a value, that value should be sent back as the request response.

```javascript
server.post('api/orderhistory',{ before: 'authFilter' } , function(request){
  var ORDERS = some data;
  return [200, "Content-Type": "application/json" , ORDERS];
});

server.filter('authFilter', function( request) {
 //chech request headers to see if access token exists otherwise return 
    return [403,{"Content-Type": "application/json"},JSON.stringify({status:'failed',message:'access denied'})];
});
```
this way you can define a filter once and enforce it on multiple routes. the key is treating a possible response from filters as response for the request.

jacobthemyth:
It's on my todo list to put together a pull request for this, but if you need to implement this in the meantime, I did the following:

```js
/*
 * Monkey patch Pretender to allow a "global" handler that returns 401 if the
 * headers are incorrect.
 */
var _handlerForOriginal = Pretender.prototype._handlerFor;
Pretender.prototype._handlerFor = function(verb, path, request){
  if( ! hasAuthHeaders(request) ) {
    return _handlerForOriginal('GET', '/unauthorized', request);
  }
  return _handlerForOriginal.apply(this, arguments);
};

var server = new Pretender(function(){
  this.get('/unauthorized', function (req) {
    return [401, {'Content-Type': 'application/json'}, {error: "unauthorized"}];
  });
});
```

PR #104, opened 2015-08-06
Added json helper to Pretender

stefanpenner:
- Added Pretender.json() helper
- Added tests for json helper
- Added docs to README about json helper

PR created on @stefanpenner's recommendation

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,14 @@ server.put('/api/songs/99', function(request){
 
 ```
 
+Pretender comes with a `json` helper that takes status code and response object and returns handler function. 
+
+```javascript
+var server = new Pretender();
+var json = Pretender.json;
+server.put('/api/songs/99', json(404, {}));
+```
+
 ### Paths
 Paths can either be hard-coded (`this.get('/api/songs/12')`) or contain dynamic segments
 (`this.get('/api/songs/:song_id'`). If there were dynamic segments of the path,
--- a/pretender.js
+++ b/pretender.js
@@ -371,6 +371,16 @@ Pretender.parseURL = parseURL;
 Pretender.Hosts = Hosts;
 Pretender.Registry = Registry;
 
+Pretender.json = function json(status, payload) {
+  return function() {
+    return [
+      status,
+      { 'Content-Type': 'text/json' },
+      JSON.stringify(payload)
+    ];
+  };
+};
+
 if (typeof module === 'object') {
   module.exports = Pretender;
 } else if (typeof define !== 'undefined') {
--- a/test/json_helper_test.js
+++ b/test/json_helper_test.js
@@ -1,0 +1,28 @@
+module('json helper');
+
+var json = Pretender.json;
+
+test('is present', function() {
+  ok(json, 'json helper is available');
+});
+
+test('accepts and returns error code', function() {
+  var success = json(200, {})();
+  var notFound = json(404, {})();
+  equal(success[0], 200, 'passing status code 200 returns 200 as status code');
+  equal(notFound[0], 404, 'passing status code 404 returns 404 as status code');
+});
+
+test('response is stringified', function() {
+  var response = { foo: 'bar' };
+  var success = json(200, response)();
+
+  notEqual(success[2], response, 'return value doesn\'t equal to passed in object');
+  equal(success[2], '{"foo":"bar"}', 'response is a string');
+});
+
+test('encoding is always json', function() {
+  var success = json(200, { foo: 'bar'});
+
+  equal(success()[1]['Content-Type'], 'text/json', 'Content Type is set to text/json');
+});


trek:
Still not a huge fan, although this is a common case. How would I add more headers when I need them?

The foundational problem here (mutating requests on the way in and and responses on the a way out) seems better handled via the middleware approach #70.

A middleware could allow an entire server to use a custom public API for return values in the responder by exposing the normal api to the framework's internals.

stefanpenner:
> How would I add more headers when I need them?

In-practice this seems rare, or could be added via a middleware/preprocess hook.

The most common use-case seems to be:

* JSON response, with some status code

before:
```js
server.get('path', function() {
  return [200, { "Content-Type': "application/json" }, JSON.stringify({})]
});
```

after:

```js
server.get('path', json(200, {}));
```

If users need custom headers on a specific request, they can peel back the JSON sugar and :shipit:

DingoEatingFuzz:
Seems like if you are going to add a json helper, you are opening the floodgates to any number of other helpers.

I also really don't like the opinion of other headers being uncommon. It's a good hunch, but it's still an opinion leaking in.

I have handled this, for the specific APIs I have to deal with, with a catchall approach:

```js
api.prepareHeaders = function(headers) {
  headers['content-type'] = 'application/json';
  return headers;
};

api.prepareBody = function(body) {
  return JSON.stringify(body || {});
};
```

I have also written a large number of helper functions for composing handlers. Here are a few.

```js
this.get('/api/account', relatedAccounts('account1'));
this.get('/api/account/:id', one('accounts'));
this.get('/api/account/all', all('accounts', 'account_id'));
this.get('/api/auth', allOrOne('auths'));
this.post('/api/auth', mustPatch());
this.delete('/api/segment/:id', noResponse());
```

The `one`, `all`, and `allOrOne` helpers all integrate with our fixtures, `mustPatch` throws if there is isn't a patch provided for the route (patching is another system we have), and `noResponse` is actually reusable. 

Here are a couple of the definitions for reusable helpers.

```js
export function empty() {
  return patchableResponder(function(req) {
    return ok({ });
  });
}

export function noResponse() {
  return patchableResponder(function(req) {
    return noContent();
  });
}
```

Here `ok` and `noContent` are helpers for building envelopes:

```js
function ok() {
  return [ 200, {} ].concat([].slice.call(arguments));
}

function noContent() {
  return [ 204, {} ];
}
```

I don't think any of this stuff should end up in Pretender, but hopefully this serves as some food for thought.

There is more than one way to go about building out these helpers, and I think those opinions should come from some other project.

trek:
If we lived in a world with great package managers I'd just say we want another library that provides common helpers for specific common cases, like authoring a json api. This can easily live within the pretenderjs org

```
import Pretender from "pretender";
import {json, ok} from "pretender-json-api";

let server = new Pretender(function(){
  this.get('/all/songs', function(req){
    let myObject = {
      key: 'val'
    };
    
    return ok(json(myObject))
  })
});
```

cibernox:
I'm going to lure @samselikoff here as someone that builds on top predender and has actually already built a DSL to make it nicer to use.

stefanpenner:
> I also really don't like the opinion of other headers being uncommon. It's a good hunch, but it's still an opinion leaking in.
> 
> I have handled this, for the specific APIs I have to deal with, with a catchall approach:

you example is also the solution i suggest

DingoEatingFuzz:
> the basic JSON case is by far the most common. Nothing prevents apps users from taking the approach and having there own helpers to cleanup code.

Definitely the most common, and it wouldn't be the end of the world if this got merged.

> opinions guide an ecosystem

Fair point, still doesn't mean they need to be in Pretender core (in a world where Pretender has a core and an ecosystem).

> your example is also the solution i suggested.

Not true. It has pieces that are similar to what you suggested, but the JSON.stringify is happening in `prepareBody`, and every one of my handlers are wrapped in a `patchableResponder`. I could rework it to use the proposed `Pretender.json`, but I wouldn't, since it doesn't gain me anything.

Specific comments aside, I agree with Trek when he says "If we lived in a world with great package managers".

It's on my todo list to move the general purpose utils I have written into its own package, but that still wouldn't be convenient for many users.

stefanpenner:
> the basic JSON case is by far the most common. Nothing prevents apps users from taking the approach and having there own helpers to cleanup code.
> Definitely the most common, and it wouldn't be the end of the world if this got merged.

then this seems like a lot of energy, to basically say \o/.

I have lost interest – @trek :+1: / :-1: and we can move on.

trek:
:-1: for including in the core, despite this also being my most common use case.

stefanpenner:
sounds fine.

trek:
:+1: on creating `pretedner-helpers` or `pretender-middlewares` especially if @samselikoff wants to generalize some things he's written

samselikoff:
would love to chat more about this! I think I'll be ready after I finish a few more key features

Issue #85, opened 2015-07-24
jQuery 1.x with passthrough


code0100fun:
I can't get passthrough to work with jQuery 1.x. The requests are sent using `$.ajax('/path').then(...)` but the `then` callbacks are never fired.

Seems like the `onload` event is detected on the browser `xhr` but is not supported by the 1.x jQuery [`xhr`](https://github.com/jquery/jquery/blob/1.11.3/src/ajax/xhr.js) module, so it is not [dispatched](https://github.com/pretenderjs/pretender/blob/053b865064cf09dc6a07800b7a5039880381a8df/pretender.js#L168) when the response is received.

I was able to get the events to fire by removing the check for [`onload`](https://github.com/pretenderjs/pretender/blob/053b865064cf09dc6a07800b7a5039880381a8df/pretender.js#L143) and just adding both events. 

There is a chance that both events could be fired I'm guessing. Any other concerns I should check into before getting a PR ready?

trek:
Seems bad. PR would be help.

kottenator:
I'm using jQuery 1.x too and it doesn't work with Pretender 0.9.0
Unfortunately I can't switch to jQuery 2.x :( 
Could you (Pretender developers), please, restore jQuery 1.x support?

olivierlesnicki:
@trek any update on this one?

olivierlesnicki:
@code0100fun do you have a PR for this one?

code0100fun:
I can't think of a clean way to tell when the user has an older version of jQuery without tying Pretender to jQuery in some way.

As a workaround, you may be able to wrap the [`onload`](https://github.com/pretenderjs/FakeXMLHttpRequest/blob/f48f3b8087ecb6df2ffd16ef31bd0e1bed7989f7/fake_xml_http_request.js#L144) callback of FakeXMLHTTPRequest and call `onreadystatechange`. That is what [jQuery 1.x](https://github.com/jquery/jquery/blob/1.11.3/src/ajax/xhr.js#L170) is using as a success callback.

[...]

PR #105, opened 2015-08-10
Fix jQuery 1.x never calling complete

code0100fun:
Fixes https://github.com/pretenderjs/pretender/issues/85.

Tried to find a way to check for this case without referring to jQuery directly but it can't really be avoided.

Tests should ideally be ran against jQuery 1.x and 2.x but that is going to require some project restructuring.

Two of the async/synchronous tests fail with jQuery 1.x for some reason. It seems like a difference in the way 1.x interacts with XMLHttpRequest. I think it may need to be addressed upstream (in FakeXMLHttpRequest). Any ideas @trek or @mike-north?

### DIFF
--- a/pretender.js
+++ b/pretender.js
@@ -7,6 +7,7 @@ var appearsBrowserified = typeof self !== 'undefined' &&
 
 var RouteRecognizer = appearsBrowserified ? require('route-recognizer') : self.RouteRecognizer;
 var FakeXMLHttpRequest = appearsBrowserified ? require('fake-xml-http-request') : self.FakeXMLHttpRequest;
+var jQuery = appearsBrowserified ? require('jquery') : self.jQuery;
 
 /**
  * parseURL - decompose a URL into its parts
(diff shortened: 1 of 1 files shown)


trek:
Can you show some of the other strategies you tried? This add jQuery as a dependency, when there isn't a reason to.

pretender.js
```diff
@@ -7,6 +7,7 @@ var appearsBrowserified = typeof self !== 'undefined' &&
 
 var RouteRecognizer = appearsBrowserified ? require('route-recognizer') : self.RouteRecognizer;
 var FakeXMLHttpRequest = appearsBrowserified ? require('fake-xml-http-request') : self.FakeXMLHttpRequest;
+var jQuery = appearsBrowserified ? require('jquery') : self.jQuery;
```
trek: This feels wrong for projects that don't have jQuery as a dependency.

mike-north: agreed

PR #108, opened 2015-08-22
Remove Timeout, WithCredential and onProgress for synchronous XHR

step2yeung:
I need pretender to send synchronous XHR on passthrough, and currently, pretender will add Timeout, WithCredential and onProgress event to the synchronous XHR.

These should be removed for synchronous XHR:
http://stackoverflow.com/questions/12472395/withcredentials-property-is-not-supported-in-firefox
https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest
Use of XMLHttpRequest's withCredentials attribute is no longer supported in the synchronous mode in window context.

https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest
You may not use a timeout for synchronous requests with an owning window.

https://bugs.webkit.org/show_bug.cgi?id=40996
Progress event should not be fired during synchronous XMLHttpRequest

On chrome, the rules below are not enforced, but it is in Firefox, so request will fail on firefox.

### DIFF
--- a/pretender.js
+++ b/pretender.js
@@ -134,7 +134,7 @@ function interceptor(pretender) {
   };
 
   // passthrough handling
-  var evts = ['error', 'timeout', 'progress', 'abort'];
+  var evts = ['error', 'timeout', 'abort'];
   var lifecycleProps = ['readyState', 'responseText', 'responseXML', 'status', 'statusText'];
   function createPassthrough(fakeXHR) {
     var xhr = fakeXHR._passthroughRequest = new pretender._nativeXMLHttpRequest();
@@ -144,6 +144,11 @@ function interceptor(pretender) {
       evts.push('load');
     } else {
       evts.push('readystatechange');
+    }
+
+    // add progress event for async calls
+    if (fakeXHR.async) {
+      evts.push('progress');
     }
 
     /*jshint -W083 */
@@ -170,8 +175,10 @@ function interceptor(pretender) {
     /*jshint +W083 */
     // jscs:enable requireCurlyBraces
     xhr.open(fakeXHR.method, fakeXHR.url, fakeXHR.async, fakeXHR.username, fakeXHR.password);
-    xhr.timeout = fakeXHR.timeout;
-    xhr.withCredentials = fakeXHR.withCredentials;
+    if (fakeXHR.async) {
+      xhr.timeout = fakeXHR.timeout;
+      xhr.withCredentials = fakeXHR.withCredentials;
+    }
     for (var h in fakeXHR.requestHeaders) {
       xhr.setRequestHeader(h, fakeXHR.requestHeaders[h]);
     }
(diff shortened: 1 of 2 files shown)


step2yeung:
Test to come..

step2yeung:
Squashed all previous commit to b579407

step2yeung:
@nathanhammond Could you help me take a look?

pretender.js
```diff
@@ -146,6 +146,11 @@ function interceptor(pretender) {
       evts.push('readystatechange');
     }
 
+    // add progress evt for async call
```
nathanhammond: evt => event, call => calls

nathanhammond:
Stephen, thanks for doing all of the research on the currently supported features for synchronous versus asynchronous XHRs.

@trek @mike-north @stefanpenner Is there anything else we need to do in order to land this?

test/passthrough_test.js
```diff
@@ -38,3 +38,84 @@ asyncTest('allows matched paths to be pass-through', function(assert) {
...
+
+asyncTest('synchronous request does not have timeout, withCredentials and onprogress event', function(assert) {
+  function testXHR() {
+    this.pretender = pretender;
+    this.open = function() {};
+    this.setRequestHeader = function() {};
+    this.send = {
+      apply: function(xhr, argument) {
+        assert.equal(xhr.timeout, undefined);
+        assert.equal(xhr.withCredentials, undefined);
```
stefanpenner: these assertions should be `!(property in xhr)`

test/passthrough_test.js
```diff
@@ -38,3 +38,84 @@ asyncTest('allows matched paths to be pass-through', function(assert) {
...
 });
+
+asyncTest('asynchronous request with pass-through has timeout, withCredentials and onprogress event', function(assert) {
+  function testXHR() {
+    this.pretender = pretender;
+    this.open = function() {};
+    this.setRequestHeader = function() {};
+    this.send = {
+      apply: function(xhr, argument) {
+        assert.equal(xhr.timeout, 1000);
```
stefanpenner: these assertions should be !(propertyName in xhr)

test/passthrough_test.js
```diff
@@ -38,3 +38,84 @@ asyncTest('allows matched paths to be pass-through', function(assert) {
...
+        assert.ok(xhr.onprogress !== undefined);
+        this.pretender.resolve(xhr);
+      }
+    };
+  }
+  pretender._nativeXMLHttpRequest = testXHR;
+
+  pretender.post('/some/path', pretender.passthrough);
+
+  $.ajax({
```
stefanpenner: is it actually important that we use $.ajax or should we just use the raw.

step2yeung: Its not really important, I just followed the other test since majority was using $.ajax.

stefanpenner: :+1:

stefanpenner:
v0.10.0 released which should include this.

step2yeung:
Awesome! Thanks @stefanpenner & @nathanhammond !

PR #111, opened 2015-09-02
Decode URI for phantomjs

step2yeung:
Due to the v 0.8 changes to add the hosts, parseURL(url).fullpath is returning different values for phantomjs and Chrome/Firefox when we have special characters are involved in the URL. 

Example:

For url = '/some/path/' + encodeURIComponent('encoded:char'), we get.. '/some/path/encoded%3Achar'
Hence we call: pretender.get( '/some/path/encoded%3Achar', somejson );

When pretender intercepts and handlesRequest
In Phantomjs, parseURL(url).fullpath = /some/path/encoded%3Achar
while in Chrome, parseURL(url).fullpath = /some/path/encoded%253Achar  (double encoding)

Also, within routeRecognizer, the states are stored differently.
In Phantomjs, registry.recognize matches for /some/path/encoded:char in the state.
while in Chrome, registry.recognize matches for /some/path/encoded%3Achar in the state.

### DIFF
--- a/pretender.js
+++ b/pretender.js
@@ -365,7 +365,11 @@ Pretender.prototype = {
   },
   _handlerFor: function(verb, url, request) {
     var registry = this.hosts.forURL(url)[verb];
-    var matches = registry.recognize(parseURL(url).fullpath);
+    var fullpath = parseURL(url).fullpath;
+    var matches = registry.recognize(fullpath);
+    if (matches === undefined) {
+      registry.recognize(decodeURIComponent(fullpath));
+    }
 
     var match = matches ? matches[0] : null;
     if (match) {
--- a/test/calling_test.js
+++ b/test/calling_test.js
@@ -457,3 +457,16 @@ test('resolves cross-origin requests', function() {
   ok(wasCalled);
 
 });
+
+test('url with encoded characters', function() {
+  var wasCalled;
+  var encodedChar = 'encoded:char';
+  var urlpath = '/some/path/' + encodeURIComponent(encodedChar);
+
+  pretender.get(urlpath, function() {
+    wasCalled = true;
+  });
+
+  $.ajax({url: urlpath});
+  ok(wasCalled);
+});


step2yeung:
@nhammond Do you mind taking a look at this?

nathanhammond:
I'm unable to reproduce the issue after trying to reduce the problem to the simplest form. These both produce the same results for me:

## Chrome ##
```javascript
function parseURL(url) {
  var anchor = document.createElement('a');
  anchor.href = url;
  anchor.fullpath = anchor.pathname + (anchor.search || '') + (anchor.hash || '');
  return anchor;
}
console.log(parseURL('/asdf%3Afdsa').fullpath);
```

## Phantom ##
```javascript
var page = require('webpage').create();
var url = 'https://github.com/';
page.open(url, function (status) {
  function parseURL(url) {
    var anchor = document.createElement('a');
    anchor.href = url;
    anchor.fullpath = anchor.pathname + (anchor.search || '') + (anchor.hash || '');
    return anchor;
  }
  console.log(parseURL('/asdf%3Afdsa').fullpath);
  phantom.exit();
});
```

@step2yeung Can you confirm versions of Chrome/Phantom that are affected and that my reproduction is checking for the correct thing?

jamesarosen:
Some notes:

 * RouteRecognizer's `recognize` [calls decodeURI](https://github.com/tildeio/route-recognizer/blob/0.1.9/lib/route-recognizer.js#L498)
 * Pretender's `parseURL` implementation means `server.get('/foo bar')` and `server.get('/foo%20bar')` will do the same thing. Both will create routes whose regex is `/^\/foo%20bar$/`

I don't have a firm theory as to why this breaks on Chrome, but we see it quite reliably on Ember 1.13.8 with the following failing test:

```js
test("You can specify URL-encoded requests", function(assert) {
  QUnit.config.testTimeout = 400;
  var done = assert.async();

  server.get('/all%20pets/', function(db) {
    return { pets: db.pets };
  });

  var request = Ember.$.ajax({
    url: '/all%20pets/'
  });

  request.then(function(response, statusText, jqXHR) {
    assert.equal(jqXHR.status, 200, 'Requests with URL-encoded characters match');
    done();
  });
});
```

pretender.js
```diff
@@ -365,7 +365,11 @@ Pretender.prototype = {
   },
   _handlerFor: function(verb, url, request) {
     var registry = this.hosts.forURL(url)[verb];
-    var matches = registry.recognize(parseURL(url).fullpath);
+    var fullpath = parseURL(url).fullpath;
+    var matches = registry.recognize(fullpath);
+    if (matches === undefined) {
+      registry.recognize(decodeURIComponent(fullpath));
```
jamesarosen: This should be `matches = ...` right?

trek:
Handled by https://github.com/pretenderjs/pretender/pull/163

PR #114, opened 2015-09-22
Update pretender.js

### LINKED ISSUE #85: jQuery 1.x with passthrough

olivierlesnicki:
fix #85

### DIFF
--- a/pretender.js
+++ b/pretender.js
@@ -142,9 +142,9 @@ function interceptor(pretender) {
     // Use onload instead of onreadystatechange if the browser supports it
     if ('onload' in xhr) {
       evts.push('load');
-    } else {
-      evts.push('readystatechange');
     }
+    
+    evts.push('readystatechange');
 
     // add progress event for async calls
     if (fakeXHR.async) {
--- a/pretender.js
+++ b/pretender.js
@@ -143,7 +143,7 @@ function interceptor(pretender) {
     if ('onload' in xhr) {
       evts.push('load');
     }
-    
+
     evts.push('readystatechange');
 
     // add progress event for async calls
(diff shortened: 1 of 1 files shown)


trek:
Can you check my CI is failing?

thijsvdanker:
CI was failing because extracting FakeXMLHttpRequest gave an error
```
bower invalid-meta  fake-xml-http-request is missing "ignore" entry in bower.json
```
And pretender.js fails when requiring 'fake-xml-http-request'.

I was having issues with callbacks not being called as well, even when using jQuery 2.x.
Using @olivierlesnicki PR fixed the issue for me, but I have too little knowledge of the internals to say if it's good.

thijsvdanker:
@trek Can you run CI again to see if it was indeed a hickup?

olivierlesnicki:
@thijsvdanker I've triggered the CI again

trek:
CI is still failing, sadly.

thijsvdanker:
This time CI actually fails because of the patch. I can reproduce the failing test with running `npm test` locally.
Reverting the patch makes the test pass again.

john-kurkowski:
Should this be closed in favor of #130? Fixes the same issue, performs refactoring, tests pass. All it needs is to add jQuery 1.x to the test suite.

olivierlesnicki:
@john-kurkowski agreed

trek:
Closing in favor of #130

PR #116, opened 2015-09-30
update bower.json version to match package.json

bantic:


bantic:
this is a dupe of #113

PR #127, opened 2015-11-15
Fix repository URL in package.json

tricknotes:
Now pretender is organized by [pretenderjs](https://github.com/pretenderjs).


rwjblue:
:+1:

PR #129, opened 2015-11-16
use sinon.fakeTimers to reduce test flakiness

fivetanley:
These tests pass intermittently due to the unreliability of using
setTimeout in a scheduled fashion. Introducing fakeTimers allow us to
more reliably manipulate in-browser timing APIs.


adamjmcgrath:
+1 I was seeing some intermittent failures for these tests when I was running them locally

stefanpenner:
@fivetanley rebase? Then we can merge

stefanpenner:
https://github.com/pretenderjs/pretender/pull/206

PR #130, opened 2015-11-18
Fix readystatechange

adamjmcgrath:
- Add `readystatechange` to xhr events regardless of the browser supporting `onload`, for ajax implementations that still rely on readystatechange (eg jQuery 1.x #85 )
- Move `evts` into function closure so multiple calls to `createPassthrough` don't result in duplicate entries in `evts`.

### DIFF
--- a/pretender.js
+++ b/pretender.js
@@ -182,8 +182,6 @@ function interceptor(pretender) {
       };
     }
 
-    xhr.open(fakeXHR.method, fakeXHR.url, fakeXHR.async, fakeXHR.username, fakeXHR.password);
-
     // set the on- handler on the native xhr's `upload` property for
     // the given eventType
     function createUploadHandler(eventType) {
@@ -193,6 +191,8 @@ function interceptor(pretender) {
         };
       }
     }
+
+    xhr.open(fakeXHR.method, fakeXHR.url, fakeXHR.async, fakeXHR.username, fakeXHR.password);
 
     var i;
     for (i = 0; i < evts.length; i++) {
--- a/pretender.js
+++ b/pretender.js
@@ -133,23 +133,22 @@ function interceptor(pretender) {
     }
   };
 
-  // event types to handle on the xhr
-  var evts = ['error', 'timeout', 'abort'];
-
-  // event types to handle on the xhr.upload
-  var uploadEvents = ['progress'];
-
-  // properties to copy from the native xhr to fake xhr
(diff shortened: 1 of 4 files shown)


trek:
If this is to fix issues with jQuery 1.x, the test suite needs to be updated to run integration tests against both jQuery 1.x and 2.x or we'll have regressions.

adamjmcgrath:
This is to fix issues with any ajax client that uses the `readystatechange` event, of which jQuery 1.x is one. 

The additions to [passthrough_test.js](/pretenderjs/pretender/pull/130/files?diff=unified#diff-68e248a96e1a40b8b9402f3349c1332bR166) are designed to stop regressions of this case.

pretender.js
```diff
@@ -201,7 +202,6 @@ function interceptor(pretender) {
       createUploadHandler(uploadEvents[i]);
     }
 
-    xhr.open(fakeXHR.method, fakeXHR.url, fakeXHR.async, fakeXHR.username, fakeXHR.password);
```
omouse: why was it necessary to move where the xhr.open is being called?

adamjmcgrath:
> why was it necessary to move where the xhr.open is being called?

Good point, I probably should have explained this. There's a bug (https://bugs.webkit.org/show_bug.cgi?id=45994#c2) in older versions of webkit (that effect phantomjs 1.x) where the status is not accessible on an `XMLHttpRequest` if the `readyState` is 0 or 1. When we call `xhr.open`, a `readystatechange` event is fired and we try to copy the status of the `xhr` to the `fakeXHR` - this throws the "DOM Exception 11" that was causing [pr#114](114)'s build to fail (https://travis-ci.org/pretenderjs/pretender/builds/90093794#L307). Moving the `xhr.open` to before the `createHandler` execution means that we don't try and copy the status until after xhr.send, when the readyState is > 1.

The downside is that the passthrough request doesn't proxy the initial `readystatechange` event. Another solution could be to upgrade phantomjs to 2.x - although getting a linux binary that works on travis might not be trivial.

Also, for neatness - I should really have moved it to below the `createUploadHandler ` definition rather than above - will do that.

> If this is to fix issues with jQuery 1.x, the test suite needs to be updated to run integration tests against both jQuery 1.x and 2.x or we'll have regressions.

Also, on retrospect, I agree - will take a look at adding tests for jQuery 1 and 2 at some point this weekend.

trek:
Looks good!

PR #132, opened 2015-11-30
FakeXMLHttpRequest in bower.json to match fake-xml-http-request in package.json

john-kurkowski:
So I can get fixes from the latest published version of FakeXMLHttpRequest.


xcambar:
:+1:

PR #134, opened 2015-12-14
pass headers to prepareBody

bekzod:
passing headers to prepareBody can be useful when you want to serialize depending `content-type` header, for example if it is `text/...`

### DIFF
--- a/pretender.js
+++ b/pretender.js
@@ -323,7 +323,7 @@ Pretender.prototype = {
 
         var status = statusHeadersAndBody[0],
             headers = this.prepareHeaders(statusHeadersAndBody[1]),
-            body = this.prepareBody(statusHeadersAndBody[2]),
+            body = this.prepareBody(statusHeadersAndBody[2], headers),
             pretender = this;
 
         this.handleResponse(request, async, function() {


DingoEatingFuzz:
Why do this when there is a `prepareHeaders` method? 

Are you needing the headers to make decisions regarding the body?

bekzod:
@DingoEatingFuzz you mean "why not to do this inside `prepareHeaders`" ? 
well there you don't have access to body 
you mean "Do you need headers to make decisions regarding body?"
yes

```javascript
   this.get('/api/songs', function(request){
    return [
      200,
      {'content-type': 'text/csv'},
      "firstname,lastname\nbob,dylon"
    ];
  });
```
and if I have `prepareBody` like this 
```javascript
server.prepareBody = function(body){
  return body ? JSON.stringify(body) : '{"error": "not found"}';
}
```
so I would like to do this 

```javascript
server.prepareBody = function(body, headers) {
  if (headers['content-type'] === 'application/javascript') {
    return body ? JSON.stringify(body) : '{"error": "not found"}';
  } else {
    return body;
  }
}
```
then I get ""firstname,lastname\nbob,dylon"" double quotes as the header is `text/csv`

trek:
Thanks!

trek:
This is now released in 0.12.0

PR #140, opened 2016-01-09
Use nested QUnit style

trek:
I've never liked the "a test is in a module if it follows a call to `module`" style QUnit:
  
  * Bumps to newer QUnit for development 
  * Refactors test suite into nested QUnit style
  * Refactors suite to use `beforeEach`/`afterEach` instead of `setup`/`teardown`
  * Removes a bunch of top-level var setting/unsetting

### DIFF
--- a/test/calling_test.js
+++ b/test/calling_test.js
@@ -1,459 +1,462 @@
-var pretender;
-module('pretender invoking', {
-  setup: function() {
-    pretender = new Pretender();
-  },
-  teardown: function() {
-    if (pretender) {
-      pretender.shutdown();
-    }
-    pretender = null;
-  }
-});
+var describe = QUnit.module;
+var it = QUnit.test;
 
-test('a mapping function is optional', function() {
-  var wasCalled;
-
-  pretender.get('/some/path', function() {
-    wasCalled = true;
+describe('pretender invoking', function(config) {
+  config.beforeEach(function() {
+    this.pretender = new Pretender();
   });
 
-  $.ajax({url: '/some/path'});
-  ok(wasCalled);
-});
+  config.afterEach(function() {
+    this.pretender.shutdown();
+  });
 
-test('mapping can be called directly', function() {
-  var wasCalled;
-  function map() {
-    this.get('/some/path', function() {
+  it('a mapping function is optional', function() {
+    var wasCalled;
+
+    this.pretender.get('/some/path', function() {
       wasCalled = true;
     });
-  }
 
-  pretender.map(map);
-
-  $.ajax({url: '/some/path'});
-  ok(wasCalled);
-});
-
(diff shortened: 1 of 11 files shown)


rwjblue:
:+1: - I love this style also

test/calling_test.js
```diff
@@ -1,459 +1,462 @@
...
-  },
-  teardown: function() {
-    if (pretender) {
-      pretender.shutdown();
-    }
-    pretender = null;
-  }
-});
+var describe = QUnit.module;
+var it = QUnit.test;
```
leobalter: I knew this was going to happen eventually after landing the nested modules. I'm glad it worked well for you.

test/calling_test.js
```diff
@@ -1,459 +1,462 @@
...
-  pretender.handledRequest = function(verb, path, request) {
-    ok(true, 'handledRequest hook was called');
-    equal(verb, 'GET');
-    equal(path, '/some/path');
-    equal(request.responseText, json);
-    equal(request.status, '200');
-    done();
-  };
+  it('handledRequest is called', function(assert) {
+    var done = assert.async();
```
leobalter: using the `assert` parameter is recommended to avoid assertion leaking to other tests on async calls. As you are using it here, you could extend it to the other tests. As in other test suites you can call it as `t` only.

This async leaking on QUnit assertions happens because QUnit tests register each assertion, which is great for logging, reports and used along with `expect`, which is great to avoid false positives.

leobalter:
@trek, I'm sorry for the noise, I've got excited seeing this work. I'm looking forward to know if you have more feedback on QUnit.

test/calling_test.js
trek: I'm not sure I understand what you're suggesting. Can you show an example?

leobalter: A few lines below you call some assertions from the global scope, like: 

```js
    this.pretender.handledRequest = function(verb, path, request) {
      ok(true, 'handledRequest hook was called');
      equal(verb, 'GET');
      equal(path, '/some/path');
      equal(request.responseText, json);
      equal(request.status, '200');
      done();
    };
```

You may use the `assert` argument in order to keep the test reference, as:

```js
    this.pretender.handledRequest = function(verb, path, request) {
      assert.ok(true, 'handledRequest hook was called');
      assert.equal(verb, 'GET');
      assert.equal(path, '/some/path');
      assert.equal(request.responseText, json);
      assert.equal(request.status, '200');
      done();
    };
```

That makes a exact reference to the current test block and in case you check the reporter or even get a failure, the assertion will reference on the test they were assigned. The global assertions try to determine the current test being executed, this is proved to not go well on some async tests.

trek: I think you're looking at the likes that were removed?

PR #141, opened 2016-01-20
added option to opt-out from tracking requests

bekzod:
in some cases it might be not preferred to track requests as they can pileup according to https://github.com/samselikoff/ember-cli-mirage/issues/509 , so added option to opt out from such functionality, not sure about implementation so need feedback

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,9 +368,13 @@ server.prepareHeaders = function(headers){
 ## Tracking Requests
 Your pretender instance will track handlers and requests on a few array properties.
 All handlers are stored on `handlers` property and incoming requests will be tracked in one of
-two properties: `handledRequests` and `unhandledRequests`. The handler is also returned from
+three properties: `handledRequests`, `unhandledRequests` and `passthroughRequests`. The handler is also returned from
 any verb function. This is useful if you want to build testing infrastructure on top of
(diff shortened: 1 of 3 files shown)


samselikoff:
Looks good, could we get two tests to verify the functionality?

DingoEatingFuzz:
Is it really the case that you either want to track all requests or no requests? Seems like a whitelist or blacklist is more desirable.

olivierlesnicki:
@DingoEatingFuzz I haven't come across any `request` for whitelisting/blacklisting. It seems the only concern people have (us included) is that these responses aren't garbage collected.

Which is a big problem when running thousand of tests that make even more calls through pretender.

Leooo:
I just can't manage to run the tests after adding anything in this file: even adding a commented line makes the tests fail on 'No `onprogress` events are fired after the the request finalizes' (and I'm not talking about linters)..

samselikoff:
Believe https://github.com/pretenderjs/pretender/pull/194 will handle this

stefanpenner:
@bekzod mind reviving this one?

- [ ] add a test
- [ ] update readme

Then we can merge + ship

bekzod:
@stefanpenner done

stefanpenner:
released as v1.5.0 🎉

PR #148, opened 2016-02-03
added support for setting responseType to array buffer

### LINKED ISSUE #72: Binary files are converted to strings
We use pretender to stub an Ember application, we have some sounds that are loaded and played while using the application. Loading and playing sounds fails because the pretender doesn't handle the xhr.responseType. We use Howler to load and play sounds, Howler sets the xhr.responseType to 'arraybuffer'. Pretender doesn't handle 'arraybuffer' which results in an error because Howler expects an arraybuffer. I've been trying all day to fix this in Pretender but without success so far.

This is how Howler loads an audiofile
https://github.com/goldfire/howler.js/blob/master/howler.js#L1230

TimvdEijnden:
I was having trouble loading binary files when using [Mirage](https://github.com/samselikoff/ember-cli-mirage) in my Ember project. It seems that pretender didn't support other responses than text.

This change allow you to set the contentType of an XHR request to arraybuffer, this results in the binary content being available using the response property. 

```js
var xhr = new XMLHttpRequest();
  xhr.open("GET", '/assets/pretender.ogg', true );
  // important because this allow to load the binary data correctly, it makes the response available which contains the arraybuffer
  xhr.responseType = 'arraybuffer';
  xhr.onload = function ()
  {
   var arrayBuffer = xhr.response;
  };
  xhr.send();
```

this PR fixes #72

### DIFF
--- a/pretender.js
+++ b/pretender.js
@@ -146,13 +146,18 @@ function interceptor(pretender) {
 
     var xhr = fakeXHR._passthroughRequest = new pretender._nativeXMLHttpRequest();
 
+    if (fakeXHR.responseType === 'arraybuffer') {
+      lifecycleProps = ['readyState', 'response', 'status', 'statusText'];
+      xhr.responseType = fakeXHR.responseType;
+    }
+
     // Use onload if the browser supports it
     if ('onload' in xhr) {
       evts.push('load');
     }
 
     // add progress event for async calls
-    if (fakeXHR.async) {
+    if (fakeXHR.async && fakeXHR.responseType !== 'arraybuffer') {
       evts.push('progress');
     }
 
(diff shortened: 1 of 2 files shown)


TimvdEijnden:
@trek Please review my PR, would be awesome if it's added! :musical_note:

deepan83:
We need this for our apps as well...so would be great if it is added!

TimvdEijnden:
I'm required to use my own branch in my project now. There isn't much activity lately. :crying_cat_face:

simonihmig:
Also using my own fork (#79), +1 for getting this merged!

trek:
Thanks! This will be in next release (coming shortly)

bhaidar:
Hi @TimvdEijnden can you share how you used the `responseType: arraybuffer` to return binary data? thanks

TimvdEijnden:
@bhaidar We used it to load sound files as in the example in the start.

PR #156, opened 2016-05-05
Avoid setting `QUnit.module.exports = Pretender`.

rwjblue:
When using `QUnit` along with `Pretender`, the UMD style that we were previously using would assign an `.exports` property to the `QUnit.module` method (which is present in QUnit 1.x as `self.module` also). This changes to check that we are actually in a node-like environment (where `exports === module.exports`) before setting `module.exports`.

### DIFF
--- a/pretender.js
+++ b/pretender.js
@@ -419,7 +419,7 @@ Pretender.parseURL = parseURL;
 Pretender.Hosts = Hosts;
 Pretender.Registry = Registry;
 
-if (typeof module === 'object') {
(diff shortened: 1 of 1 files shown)


Turbo87:
@rwjblue is this PR still relevant? from my recent journeys it seems that the current Pretender releases don't have any issues with QUnit running in parallel anymore

rwjblue:
@Turbo87 - It is relevant when using QUnit 1.x, but not 2.x (since 2.x doesn't set `module` global).

PR #162, opened 2016-07-15
Fix for passthrough requests with empty responseText

adamjmcgrath:
There's a somewhat edge case issue we encountered where blank strings returned from the passthrough request weren't being copied on to the fake xhr's responseText property, eg:

``` javascript
// server
app.get('/foo', function() {
  return '';
});

// client
xhr.open('GET', '/foo');
xhr.onload = function() {
  assert(xhr.responseText === ''); // Error (xhr.responseText === null)
}
xhr.send();
```

### DIFF
--- a/pretender.js
+++ b/pretender.js
@@ -165,7 +165,7 @@ function interceptor(pretender) {
     function copyLifecycleProperties(propertyNames, fromXHR, toXHR) {
       for (var i = 0; i < propertyNames.length; i++) {
         var prop = propertyNames[i];
-        if (fromXHR[prop]) {
+        if (fromXHR.hasOwnProperty(prop)) {
           toXHR[prop] = fromXHR[prop];
         }
       }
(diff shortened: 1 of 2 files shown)


pretender.js
```diff
@@ -165,7 +165,7 @@ function interceptor(pretender) {
     function copyLifecycleProperties(propertyNames, fromXHR, toXHR) {
       for (var i = 0; i < propertyNames.length; i++) {
         var prop = propertyNames[i];
-        if (fromXHR[prop]) {
+        if (fromXHR.hasOwnProperty(prop)) {
```
stefanpenner: this should likely be `prop in fromXHR` or the semantics will change unexpectedly.

stefanpenner: https://github.com/pretenderjs/pretender/pull/201

stefanpenner:
fixup + merge #201 here, thanks!

Issue #159, opened 2016-05-27
support requests containing form data


jbailey4:
How should requests with form data be handled? Is there way to parse form data with pretender?
Currently in the FakeRequest containing form data the request body looks like the following: 

`requestBody: "foo=bar&baz=someValue"`

dustinfarris:
i am curious about this as well

crodriguez1a:
@jbailey4 I'm also running into a similar issue. Any traction on this?

trek:
Parsing form data into language-specific objects is not handled identically across application servers. At a cursory investigation a while back there were so many small differences that it didn't seem reliable to include this in Pretender without spawning a list of Issues related to "parses X wrong" (where "wrong" means "not like the server application library I am communicating with in production").

Indeed, just for nodejs, there are several competing body parsers that all return slightly different outputs for url encoded form bodies: https://github.com/expressjs/body-parser#bodyparserurlencodedoptions, https://www.npmjs.com/package/body#readme, https://github.com/Raynos/body, https://www.npmjs.com/package/co-body#readme

I'd be happy to recommend in the README a set of HTML encoded parsers that match the parsers people are using!

jbailey4:
@trek Agreed. I think adding some docs on this in the README and mentioning parsing request bodies should be handled according to your environment setup will be helpful.

PR #155, opened 2016-04-25
issue/153 update route-recognizer verrsion in bower

ryanostrom:
Signed-off-by: Ryan Ostrom <ryan.ostrom@praxent.com>

Issue #153, opened 2016-04-20
Dependency version: route-recognizer fails on build


ryanostrom:
As a developer I want pretender's route-recognizer dependency in it's bower.json to upgrade from `0.1.1` to `0.1.9`.

**Given** I install pretender `1.1.0` to an Ember 2.3 project 
**When** I execute Ember build
**Then** I expect to the build to run without error
**Instead** I encounter the error described in [this thread](https://github.com/tildeio/route-recognizer/issues/44).

ryanostrom:
#155

trek:
Handled by https://github.com/pretenderjs/pretender/pull/163

Issue #124, opened 2015-10-29
Paths with URL-encoded characters fail to match


jamesarosen:
Given application code that makes `GET /foos/my%20foo`,

```
// This matches and request.params.name is "my foo"
this.get('/foos/:name', function(request){
  const foo =  JSON.stringify({ name: request.params.name });
  return [ 200, {"Content-Type": "application/json"}, foo ];
});

// But this never matches:
this.get('/foos/my foo', function(request) {
  ...
});

// Nor does this:
this.get('/foos/my%20foo', function(request) {
  ...
});
```

Note: I've only tested this with [Mirage](https://github.com/samselikoff/ember-cli-mirage), so its _possible_ the fault lies there. I don't see any tests for URL-encoded matches in this project, though.

jamesarosen:
Possibly related to #111 as this behaves differently on Chrome and PhantomJS

jamesarosen:
I changed `_handlerFor` to
```js
_handlerFor: function(verb, url, request){
    console.log(verb + '  ' + url);
    var registry = this.hosts.forURL(url)[verb];
    var matches = registry.recognize(parseURL(url).fullpath);

    var match = matches ? matches[0] : null;
    if (match) {
      console.log('params: ' + JSON.stringify(match.params));
      request.params = match.params;
      request.queryParams = matches.queryParams;
    }

    return match;
  },
```
and got the following output:
```
PUT /api/service/service1234/version/1/request_settings/my%20request%20setting%20%3F%20%235
params: {"serviceId":"service1234","versionNumber":"1","name":"my request setting "}
```
but it should be
```
PUT /api/service/service1234/version/1/request_settings/my%20request%20setting%20%3F%20%235
params: {"serviceId":"service1234","versionNumber":"1","name":"my request setting ? #5"}
```
or at least
```
PUT /api/service/service1234/version/1/request_settings/my%20request%20setting%20%3F%20%235
params: {"serviceId":"service1234","versionNumber":"1","name":"my request setting %3f %235"}
```
if this library doesn't want to decode params. Regardless, it shouldn't *strip* part of a param.

jamesarosen:
Failing test:
```js
test('URL-encoded params are decoded', function() {
  var params;
  pretender.get('/some/:x/:y/:z', function(request) {
    params = request.params;
  })

  $.ajax({url: '/some/a%20b/c%23d/e%3ff'});
  equal(params.x, 'a b');
  equal(params.y, 'c#d');
[...]

PR #163, opened 2016-08-20
Update Route Recognizer

nathanhammond:
The ^0.1.0 series is broken beginning after v0.1.5. This updates route-recognizer to a version with more-correct escaping semantics without regressions in specificity handling.


nathanhammond:
Fixes, closes, or supersedes:
- #32
- #124 
- #153 
- #155 

Further, #159 can be closed as that will not be supported inside of `route-recognizer`.

PR #164, opened 2016-08-24
Remove onprogress for sync requests

zglagola:
We've found that sync calls hang with onprogress events, which isn't w3c compliant. Removing these events for sync calls.

### DIFF
--- a/pretender.js
+++ b/pretender.js
@@ -139,7 +139,7 @@ function interceptor(pretender) {
     var evts = ['error', 'timeout', 'abort', 'readystatechange'];
 
     // event types to handle on the xhr.upload
-    var uploadEvents = [''];
+    var uploadEvents = [];
 
     // properties to copy from the native xhr to fake xhr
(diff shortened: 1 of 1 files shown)


zglagola:
@trek thoughts on this?

stefanpenner:
@zglagola can you share more? Isn't this interacting with the FakeXhr i'm confused how onprogress could affect anything. What am i missing.

zglagola:
@stefanpenner the inspiration came from this old webkits bug https://bugs.webkit.org/show_bug.cgi?id=40996 .

From my understanding, since this logic controls passthrough requests it wouldn't interact with FakeXhr?

That post led me to believe that the logic was wrong. On a use case that previously hung, this change worked.

stefanpenner:
@zglagola ah, i didn't notice this was the pass-through code.

stefanpenner:
released as:

```
 s/p/pretender ╍ npm publish                                                                               (2s 760ms)
+ pretender@1.4.2
```

PR #165, opened 2016-08-25
Warn on second new Pretender

trek:
Not everyone is diligent about shutting down their Pretenders. This will prevent a second instance from starting up.

### DIFF
--- a/pretender.js
+++ b/pretender.js
@@ -99,7 +99,7 @@ function Pretender(/* routeMap1, routeMap2, ...*/) {
 
   // capture xhr requests, channeling them into
   // the route map.
-  self.XMLHttpRequest = interceptor(this);
+  self.XMLHttpRequest = interceptor(this, this._nativeXMLHttpRequest);
 
   // 'start' the server
   this.running = true;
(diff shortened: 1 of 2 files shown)


fotinakis:
❤️  LGTM!

pretender.js
```diff
@@ -237,6 +237,11 @@ function interceptor(pretender) {
   };
 
   FakeRequest.prototype = proto;
+
+  if (nativeRequest.prototype._passthroughCheck) {
+    throw new Error('You created a second Pretender instance while there already one running. ' +
```
rwjblue: The text here reads a tad odd to me, maybe:

> You created a second Pretender instance while there was already one running.

rwjblue:
I'm not 100% familiar with what happens today if you do this.  If things continue to work in this scenario we should change this to a warning (or bump majors).

trek:
@rwjblue If you start new instances while a previous one is running, all new xhr requests go to the most recently created instance, which leads to unexpected routing. If you lose track of the first one you started, you can never restore the native XMLHttpRequest object.

Happy to bump majors though to avoid breaking.

fotinakis:
👍  to the major bump, makes sense.

pretender.js
trek: Suggestions?

fotinakis: He's saying that the "while there already one running" needs a "was" in there.

trek:
Closing to do a small fixup.

PR #178, opened 2016-10-06
Warn on second pretender

trek:
Replaces #156 with a small change.

### DIFF
--- a/pretender.js
+++ b/pretender.js
@@ -99,7 +99,7 @@ function Pretender(/* routeMap1, routeMap2, ...*/) {
 
   // capture xhr requests, channeling them into
   // the route map.
-  self.XMLHttpRequest = interceptor(this);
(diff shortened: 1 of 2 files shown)


rwjblue:
@trek - This was released without a major version bump (as discussed in https://github.com/pretenderjs/pretender/pull/165#issuecomment-242478399) in 1.4.0... 😿

mariogintili:
@trek Shouldn't the warning be a `console.warn`? throwing an error blocks users altogether 😢

trek:
@mariogintili @rwjblue I'm down to do a console.warn instead of throw an error release. I opted not to major bump because certain parts of Pretender (proxy and shutdown) will not work if you start a second server before shutting the first one down. 

It _was_ the existing behavior but it's also a bug and doesn't follow the [usage requirement we've had since the first release](https://github.com/pretenderjs/pretender#clean-up).

I didn't know so many project still relied on this specific bug being left in place.

PR #181, opened 2016-10-06
Remove node 0.12 for tests

trek:
Perhaps pulling the trigger a bit early, but we only really support this for running tests and that seems silly for a build about to exit LTS.


rwjblue:
Seems ok to me.

Issue #123, opened 2015-10-28
Pretender doesn't simulate empty body responses


trek:
This can happen:

```
    this.get(`/api/conversations/${conversationId}/mark_read`, function(){
      return [200, {'Content-Type': 'application/json'}, null];
    });
```

fraziern:
Hi, I'm looking at this. However on my unaltered clone, running `npm test` results in one failure:

```
...................08 10 2016 06:23:49.405:INFO [PhantomJS 1.9.8 (Mac OS X 0.0.0)]: Connected on socket WUSHb6RnzuGNJ48zAAAB with id 53549556
...............................08 10 2016 06:23:51.595:WARN [web-server]: 404: /some/path
.................08 10 2016 06:23:51.621:WARN [web-server]: 404: /some/path
....08 10 2016 06:23:51.693:WARN [web-server]: 404: /some/otherpath
.08 10 2016 06:23:51.714:WARN [web-server]: 404: /some/otherpath

PhantomJS 1.9.8 (Mac OS X 0.0.0) pretender invoking `onprogress` download events don't keep firing once the request has ended FAILED
	No `onprogress` events are fired after the the request finalizes
	Expected: 4
	Actual: 3
	    at /Users/Nick/Sites/pretender/node_modules/qunitjs/qunit/qunit.js:1485
	    at /Users/Nick/Sites/pretender/node_modules/qunitjs/qunit/qunit.js:2194
	    at /Users/Nick/Sites/pretender/test/calling_test.js:423
	
............
Chrome 53.0.2785 (Mac OS X 10.10.5): Executed 61 of 61 SUCCESS (2.767 secs / 2.751 secs)
...................08 10 2016 06:23:52.014:WARN [web-server]: 404: /some/path
08 10 2016 06:23:52.035:WARN [web-server]: 404: /some/path
.....08 10 2016 06:23:52.103:WARN [web-server]: 404: /some/otherpath
.08 10 2016 06:23:52.123:WARN [web-server]: 404: /some/otherpath
............
PhantomJS 1.9.8 (Mac OS X 0.0.0): Executed 61 of 61 (1 FAILED) (2.641 secs / 2.617 secs)
TOTAL: 1 FAILED, 121 SUCCESS
```

Any idea what's going on here?

PR #182, opened 2016-10-08
Add tests for simulating empty body response

fraziern:
For issue #123 - it appears that null and undefined body responses are already handled OK by pretender. I've added tests for this, and an extra line in the README per the issue comments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,20 @@ var server = new Pretender(function() {
 });
 ```
 
+You may return `null` or `undefined` instead of a string body, to simulate real-world responses:
+
+```javascript
+var server = new Pretender(function(){
+  this.get('/photos/:id/mark_favorite', function(request) {
+    return [
+      200,
+      {'content-type': 'application/json'},
+      null
+    ];
+  });
+});
+```
+
 ### Pass-Through
 You can specify paths that should be ignored by pretender and made as real XHR requests.
 Enable these by specifying pass-through routes with `pretender.passthrough`:
--- a/test/calling_test.js
+++ b/test/calling_test.js
@@ -481,5 +481,34 @@ describe('pretender invoking', function(config) {
 
     $.ajax({url: '/some/path'});
   });
-});
 
+  it('can return a null body response', function(assert) {
+    var done = assert.async();
+
+    this.pretender.get('/some/path', function(req) {
+      return [200, {'Content-Type': 'application/json'}, null];
+    });
+
+    this.pretender.handledRequest = function(verb, path, request) {
+      equal(request.responseText, '');
+      done();
+    };
+
+    $.ajax({url: '/some/path'});
+  });
+
+  it('can return an undefined body response', function(assert) {
+    var done = assert.async();
+
+    this.pretender.get('/some/path', function(req) {
+      return [200, {'Content-Type': 'application/json'}, undefined];
+    });
+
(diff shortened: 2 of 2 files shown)


trek:
@fraziern can you verify that this is what actually happens? My understanding was that XHR responses for no bodies had null (or maybe undefined) response texts, not empty strings?

fraziern:
@trek good question. I did some testing and it looks like request.response is undefined, while request.responseText === ''. So adding `strictEqual(request.response, undefined);` and `strictEqual(request.responseText, '');` to the tests doesn't break them.

If this is causing issues elsewhere, I can take a look, or I can just add the request.response check to the tests.

trek:
Weird, I just did a small test:

```
var express = require('express');
var app = express();

app.get('/', function(req, res){
  res.send('<html></html>')
});

app.get('/notfound', function(req, res){
  console.log("1")
  res.status(404).end();
});

app.listen(3001);

/*
  httpRequest = new XMLHttpRequest();
  httpRequest.open('GET', 'http://localhost:3000/404', true);
  httpRequest.send(null);
  httpRequest.onreadystatechange = function(){
    if (httpRequest.readyState === XMLHttpRequest.DONE) {
      debugger;
    }
  }

*/

```

and both `httpRequest.response` and `httpRequest.responseText` are `""`. What browser did you test in?

fraziern:
I was testing with the existing test framework, rather than a separate setup. 

I was able to replicate your results in Chrome. However, if I change the responseType to ```json``` first, it returns the results I got using only QUnit asserts.

Server:
```
// app.js
var express = require('express');
var app = express();
var path = require('path');

app.get('/', function(req, res) {
    res.sendFile(path.join(__dirname + '/index.html'));
});


app.get('/save', function(req, res){
  console.log("1");
  res.type('json').status(200).send().end();
});

app.listen(3000, function () {
  console.log('Example app listening on port 3000!');
});
```

Client:
```
// index.html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <title></title>
  </head>
  <body>
Testing 1, 2, 3...
  </body>

  <script type="text/javascript">
  xhr = new XMLHttpRequest();
  xhr.responseType = 'json';
  xhr.open('GET', 'http://localhost:3000/save', true);
  xhr.send(null);
  xhr.onreadystatechange = function(){
  if (xhr.readyState === XMLHttpRequest.DONE) {
    debugger;
  }
}
  </script>
</html>
```

Taking a look at the [XMLHttpRequest spec](https://xhr.spec.whatwg.org/#the-response-attribute), I think this is the correct result:

> The response attribute must return the result of running these steps:
> 
> *If responseType is the empty string or "text"
> If state is not loading or done, return the empty string.*
> 
> Return the text response.
> 
> Otherwise
> If state is not done, return null.
> 
> If responseType is "arraybuffer"
> Return the arraybuffer response.
> 
> If responseType is "blob"
> Return the blob response.
> 
> If responseType is "document"
> Return the document response.
> 
> If responseType is "json"
> Return the JSON response.

So I think where that leaves us, is we'd want ```pretender``` to properly mimic this behavior, yes? I'm taking a look at this further.

stefanpenner:
> so I think where that leaves us, is we'd want pretender to properly mimic this behavior, yes?

ya, we should mimic spec behavior

PR #194, opened 2016-12-22
don’t leak passthroughRequests

stefanpenner:


stefanpenner:
@mike-north / @trek  r?

samselikoff:
Wondering if we should add the config option, this has been requested for all tracked requests: https://github.com/pretenderjs/pretender/pull/141/files

stefanpenner:
@samselikoff can you share when the above array is used? I'm not familiar with usages, but i could imagine it is.

samselikoff:
@stefanpenner I don't think it is (someone could always be diving in), my point was just that we should add the option for the other types too since they were causing memory problems for mirage users. We can merge this in though.

stefanpenner:
@samselikoff oh, I see. You raise a good point. Thoughts on maybe making it opt-int for those arrays ?

samselikoff:
yep opt-in sounds right

trek:
I think we need to keep this as-is unless we want to do a version bump. This doesn't "leak" in the sense that Pretender _is_ intended to track these during a run. I could see adding behavior in `shutdown` if the concern is that a reference to a pretender instance remains in some other code and so references to these requests are also retained.

Correct me if I'm wrong, but it seems the underlying issues is tension between how often Pretender was intended to refresh state (before/after each test) and how Mirage would like to use Pretender (long-running, and thus susceptible to an every-growing set of bookkeeping objects).

If _that_ is the case, I think we need allow, as you suggest, the _option_ to track requests. For the current version it would need to be opt-out, but we could move to opt-in for a future version, depending how what seems to be the most typical case.

samselikoff:
Yes - I agree, and you're correct we should make it opt-out, to preserve Pretender's existing behavior and introduce no breaking changes.

stefanpenner:
> Correct me if I'm wrong, but it seems the underlying issues is tension between how often Pretender was intended to refresh state (before/after each test) and how Mirage would like to use Pretender (long-running, and thus susceptible to an every-growing set of bookkeeping objects).

Yup.



> If that is the case, I think we need allow, as you suggest, the option to track requests. For the current version it would need to be opt-out, but we could move to opt-in for a future version, depending how what seems to be the most typical case.

It seems like opt-out is fine, is it worth changing it as a default in the future or can something like mirage always flip the switch correctly?

samselikoff:
@stefanpenner want some help with this?

stefanpenner:
I think the caller of pretender should recycle pretender instance between tests. I don't think this needs to change.

PR #197, opened 2017-02-17
fixes #196 Move open before manipulating responseType to fix IE 11 bug.

### LINKED ISSUE #196: InvalidStateError on IE 11 when changing response type
Hello,

I have defined route as passthrough. On ie 11 when i fire action and this route is called i got error invalidStateError when it tries to set responseType (https://github.com/pretenderjs/pretender/blob/master/pretender.js#L173)

Check microsoft bug report: https://connect.microsoft.com/IE/feedback/details/795580/ie11-xmlhttprequest-incorrectly-throws-invalidstateerror-when-setting-responsetype

Maybe you could do some workaround and fix it? Fix for this from what i have read would be to call open before you change response type.

pawepaw:


stefanpenner:
fixing hinting: https://github.com/pretenderjs/pretender/pull/202

PR #208, opened 2017-07-20
Support running pretender in node with jsdom

### LINKED ISSUE #152: importing Pretender with transpiled es6 fails appearsBrowserified test
My test suite is transpiled from es6 using babel and I wanted to import Pretender. Calling `import Pretender from 'pretender'` and instantiating a Pretender object gave me the error:

```
TypeError: FakeXMLHttpRequest is not a function
      at interceptor (node_modules/pretender/pretender.js:119:15)
      at Object.Pretender (node_modules/pretender/pretender.js:102:25)
      at Context.<anonymous> (api-test.js:16:14)
```

I've tracked down the issue to the check for `appearsBrowserified` at the beginning of the closure in pretender.js. The final condition checks if the process is an `[object Object]` but if I console out `Object.prototype.toString.call(process)` I get `[object process]`.

I'm running node 5.5.0 with pretender 1.1.0. Is Pretender compatible with newer node versions?

### LINKED ISSUE #191: Running pretender with Jsdom errors + doesn't get assigned to XMLHttpRequest
I'm running jsdom with pretender for my integration tests. I've discovered two problems with it. The first, possibly related to #152 I believe, is the error `TypeError: FakeXMLHttpRequest is not a constructor`. I was able to fix this by changing

``` js
var appearsBrowserified = typeof self !== 'undefined' &&
                          typeof process !== 'undefined' &&
                          Object.prototype.toString.call(process) === '[object Object]';
```

to

``` js
var appearsBrowserified = typeof self !== 'undefined' &&
                          typeof process !== 'undefined' &&
                          (Object.prototype.toString.call(process) === '[object Object]' ||
                           Object.prototype.toString.call(process) === '[object process]');
```

After that, the problem was that pretender wasn't properly attaching to XMLHttpRequest (and therefore wasn't intercepting my requests. I think the problem (which may ultimately be a jsdom problem) is that the value for `self` that gets passed in to the closure (which I understand should be `window`) is different than the global `window` for jsdom. Which means that when overriding `self.XMLHttpRequest`, `window.XMLHttpRequest` doesn't get overridden. I was able to fix this by replacing

``` js
self.XMLHttpRequest = interceptor(this, this._nativeXMLHttpRequest);
```

with

``` js
self.XMLHttpRequest = interceptor(this, this._nativeXMLHttpRequest);
if(process.env.NODE_ENV === 'test') global.XMLHttpRequest = window.XMLHttpRequest = self.XMLHttpRequest;
```

I'd be happy to submit a PR, but I'm not sure about the first solution, and the second one seems hacky. Just thought I'd see what your thoughts were on this.

JakeDluhy:
Testing `Object.prototype.toString.call(process)` in several versions of node (back to `0.12.18`) I always saw it equal `[object process]`, not `[object Object]`. This is all that's preventing me from running my integration suite in node using `jsdom`.

Fixes #152 
Fixes #191

### DIFF
--- a/pretender.js
+++ b/pretender.js
@@ -3,7 +3,8 @@
 
 var appearsBrowserified = typeof self !== 'undefined' &&
                           typeof process !== 'undefined' &&
-                          Object.prototype.toString.call(process) === '[object Object]';
+                          (Object.prototype.toString.call(process) === '[object Object]' ||
+                           Object.prototype.toString.call(process) === '[object process]');
 
 var RouteRecognizer = appearsBrowserified ? require('route-recognizer') : self.RouteRecognizer;
 var FakeXMLHttpRequest = appearsBrowserified ? require('fake-xml-http-request') : self.FakeXMLHttpRequest;


padcom:
Is there any way this would get merged? Resorting to headless browser just to be able to do ajax when everything else works like a charm seems a bit excessive...

padcom:
... a month later, still the same status... Is it ever going to change?

TeknoFiend:
I've run into this issue as well.  Is this being considered?

TeknoFiend:
Could we get a comment on this?  I understand time is limited but just a quick note about whether maintainer(s) consider this a problem and/or whether the proposed solution makes any sense would be great.

samselikoff:
@trek what do you think? Seems like a simple enough change

trek:
@samselikoff I'm down. Merge away.

samselikoff:
@JakeDluhy thank you and sorry about the delay!

If you (or anyone) has time/energy to add some Node tests to the lib to cover this, that would be awesome

TeknoFiend:
Thanks for getting to this!

padcom:
Finally! Thank you! When is this going to be officially released?

samselikoff:
@padcom Believe this is in v2.1.0

PR #209, opened 2017-07-20
Enable building with webpack

### LINKED ISSUE #207: When importing Pretender from a webpack build, RouteRecognizer constructor lookup fails.
When these lines run in a webpack build, https://github.com/pretenderjs/pretender/blob/master/pretender.js#L8-L9

`RouteRecognizer` and `FakeXMLHttpRequest` point to the module, not the `default` export.

Monkey patching it with this code makes it work:

```js
  RouteRecognizer = RouteRecognizer.default || RouteRecognizer;
  FakeXMLHttpRequest = FakeXMLHttpRequest.default || FakeXMLHttpRequest;
```

Not sure what the long term solution is though.

cowboyd:
When building with WebPack, `require('x')` returns the module, not the default export, so this just peels off the default export from the module, if it's present.

fixes #207

### DIFF
--- a/pretender.js
+++ b/pretender.js
@@ -8,6 +8,11 @@ var appearsBrowserified = typeof self !== 'undefined' &&
 var RouteRecognizer = appearsBrowserified ? require('route-recognizer') : self.RouteRecognizer;
 var FakeXMLHttpRequest = appearsBrowserified ? require('fake-xml-http-request') : self.FakeXMLHttpRequest;
 
+// When building with WebPack, `require('x') returns the module, not the default export.`
+RouteRecognizer = RouteRecognizer.default || RouteRecognizer;
+FakeXMLHttpRequest = FakeXMLHttpRequest.default || FakeXMLHttpRequest;
+
+
 /**
  * parseURL - decompose a URL into its parts
  * @param  {String} url a URL


ohcibi:
@cowboyd I used this commit in an angular-cli project. The problem is that the self executing is somehow executed twice. In the first call the RouteRecognizer or FakeXMLHttpRequest do not even resolve (i.e. they are undefined) leading to another error about `.default` not being readable from `undefined`.

cowboyd:
@ohcibi I assume it doesn't work without this commit either? Can you get it working by checking if `RouteRecognizer` is defined yet?

ohcibi:
@cowboyd yep, thats exactly the quickfix I have added. You can see the debugger hitting that line twice but at the second time `RouteRecognizer` is not undefined and your fix comes to action.

I have added that into my real world project to see if it works there that way. The interesting part is that I did not had any error like that there. I was doing it very late yesterday, so I need to investigate a bit more  what was happening there (cant do it right now). But I have a clue: In the test project I messed around with the `scripts` property in the `.angular-cli.json` by putting the path to pretender into it and found out that this is not necessary when you `import` that thing directly (i'm still an angular noob). Its very likely that I did not have anything in my `scripts` in the real world project and that this error doesn't happen when one imports pretender normally via an import statement. The more I write this the more convinced I am that this is the problem. Will test it asap.

ohcibi:
Yep. Nevermind. The fix works as intended and I was misusing angular-cli.

cowboyd:
@trek @stefanpenner Any thoughts on what it will take for this to be kosher to merge? We've been using this patch for over a month without problems, but that's just one data point. I'm not sure if there's a good way to test this, short of setting up a webpack build for it.

Also, it looks like [this isn't the only effort](https://github.com/pretenderjs/pretender/pull/208) to make module lookup work in different environments, so piling on another hack might not be the right answer. Then again...

cowboyd:
Made redundant by #216 🎉

PR #212, opened 2017-09-01
Add ForcePassThrough Option

jdenquin:
Add a new option for server setup called `forcePassthrough`.
Add an other option called `disableUnhandled`.

The idea is to passthrough all request even if they are unhandled.
With this new option, we can run our test (or dev) on a real world API without register all our routes.

```javascript
var server = new Pretender({ forcePassthrough: true })
var server = new Pretender({ disableUnhandled: true })
```

The second option will simply avoid to trigger `unhandledRequest` method when a route have no handler.

I came to this cause I needed to record all my API payload to setup a VCR system. (PM me for more informations, its currently for a "private" usage but I can push and example)
Its very simple, at the end of your tests run, you just have to check your `passthroughRequests` and save it into a `json` file.

@stefanpenner

### DIFF
--- a/pretender.js
+++ b/pretender.js
@@ -121,7 +121,7 @@ function Pretender(/* routeMap1, routeMap2, ..., options*/) {
   this.unhandledRequests = shouldNotTrack ? noopArray: [];
   this.requestReferences = [];
   this.forcePassthrough = options && (options.forcePassthrough === true);
-  this.disabledUnhandled = options && (options.disabledUnhandled === true);
+  this.disableUnhandled = options && (options.disableUnhandled === true);
 
   // reference the native XMLHttpRequest object so
   // it can be restored later
@@ -340,12 +340,7 @@ Pretender.prototype = {
     var recognized = this.hosts.forURL(request.url)[verb].recognize(path);
     var match = recognized && recognized[0];
 
-    if (!match && this.forcePassthrough) {
-      this.register(verb, request.url, this.passthrough);
-      match = { handler: PASSTHROUGH };
-    }
-
-    if (match && match.handler === PASSTHROUGH) {
+    if ((match && match.handler === PASSTHROUGH) || this.forcePassthrough)  {
       this.passthroughRequests.push(request);
       this.passthroughRequest(verb, path, request);
       return true;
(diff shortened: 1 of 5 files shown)


pretender.js
```diff
@@ -154,11 +155,12 @@ function interceptor(pretender, nativeRequest) {
     }
 
     FakeXMLHttpRequest.prototype.send.apply(this, arguments);
-    if (!pretender.checkPassthrough(this)) {
-      pretender.handleRequest(this);
-    } else {
+
```
dcyriller: This is the same, isn't it ?

jdenquin: yes, I think its just better to test the truth and fallback in the else after, wdyt?

pretender.js
```diff
@@ -333,13 +335,15 @@ Pretender.prototype = {
   passthrough: PASSTHROUGH,
   checkPassthrough: function checkPassthrough(request) {
     var verb = request.method.toUpperCase();
-
     var path = parseURL(request.url).fullpath;
-
-    verb = verb.toUpperCase();
```
dcyriller: This was weird. Do you see a reason for that ?

jdenquin: certainly an oversight

dcyriller:
Do you think this option has a chance to be merged in the original repo ?

stefanpenner:
> setup a VCR system.

nice!

pretender.js
```diff
@@ -333,13 +335,15 @@ Pretender.prototype = {
...
-
     var recognized = this.hosts.forURL(request.url)[verb].recognize(path);
     var match = recognized && recognized[0];
+
+    if (!match && this.forcePassthrough) {
```
stefanpenner: if we are forcing passthrough, why does the result of `match` matter?

stefanpenner: Also when forcing passthrough, why do we register a new route? My intuition would be to immediately invoke:

```js
this.passthroughRequests.push(request);
this.passthroughRequest(verb, path, request);
```

jdenquin: just to avoid possible duplicate, I look at match, thats why i though register the route was a good idea, I give a try to your solution and come back to you!

jdenquin:
nice catch, running great, just added an other option!

stefanpenner:
looks good, but CI appears unhappy

jdenquin:
looks like a problem with Travis conf, `npm install` does not run

stefanpenner:
https://github.com/pretenderjs/pretender/pull/214

stefanpenner:
released as v1.6.0 🎉

PR #218, opened 2017-11-08
Workaround to Chrome Memory Leak

krisselden:
Old pretenders are retained after each shutdown().  Create a reference object we can detach instead of closing over pretender directly.

https://bugs.chromium.org/p/v8/issues/detail?id=4578


stefanpenner:
released as v1.6.1 🎉

PR #219, opened 2017-11-16
Use `url-parse` to parse URL & handle case when `self` is undefined

Robdel12:
## What is this?

This helps with using Pretender in a React Native project.

### DIFF
--- a/pretender.js
+++ b/pretender.js
@@ -31,9 +31,15 @@ var FakeXMLHttpRequest = appearsBrowserified ? getModuleDefault(require('fake-xm
  * }
  */
 function parseURL(url) {
-  // TODO: something for when document isn't present... #yolo
-  var anchor = document.createElement('a');
-  anchor.href = url;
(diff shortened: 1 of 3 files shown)


pretender.js
```diff
@@ -489,4 +495,4 @@ if (typeof module === 'object') {
   });
 }
 self.Pretender = Pretender;
-}(self));
+}(typeof self !== 'undefined' ? self : this));
```
rwjblue: We should use something more robust (and likely call it `global`).

For example, I use something like this in a few projects:

```js
(function(global) {

}(() => {
  if (typeof self !== 'undefined') {
    return self;
  } else if (typeof window !== 'undefined') {
    return window;
  } else if (typeof global !== 'undefined') {
    return global;
  } else {
    return Function('return this')();
  }
})());
```

Robdel12:
I'm going to close this mostly because I won't have time to come back to this 😬

PR #225, opened 2018-01-03
Assume not browserified if self.RouteRecognizer is undefined

fuse-mars:
* Issues
    * *Pretender* breaks with `Cannot read property 'prototype' of undefined` when using it with `jest` and `axios` .

* Fix
   * check to see if `self.RouteRecognizer` is defined, if not require it
   * check to see if `self.FakeXMLHttpRequest` is defined, if not require it

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,9 +1,15 @@
 {
   "name": "pretender",
-  "version": "1.5.1",
+  "version": "1.6.1",
   "lockfileVersion": 1,
   "requires": true,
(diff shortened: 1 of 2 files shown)


TeknoFiend:
Looks like this is a different solution to the same problem solved in: https://github.com/pretenderjs/pretender/pull/208

fuse-mars:
Thx @TeknoFiend, looks like it! I'll close this in favor of #208

PR #230, opened 2018-05-23
fetch support

### LINKED ISSUE #60: Using Pretender with `fetch`.
It would be nice to be able to use `fetch` instead of `XMLHttpRequest`. When using the non-native `fetch` polyfil Pretender would likely still work (because the polyfill uses `XMLHttpRequest` to implement `fetch`), but it doesn't seem that native fetch requests would be handled by Pretender.

Spec: https://fetch.spec.whatwg.org

This issue is mostly to lay out a path forward.

/cc @cowboyd @toranb

xg-wang:
tl;dr
- This commit adds support for fetch, fix #60.
- pretender swap native fetch related API if exists
- pretender.shutdown() restore native fetch related API
- doesn't work with AbortController

Changes
------
Include a fetch ponyfill and swap the native fetch during pretender
creation, then restore them when `shutdown`. Since fetch polyfill uses
xhr behind the scene, pretender should "just work".

Caveats
------
1. The supplement set of the polyfill implementation with the spec includes (not complete):
- Inability to [set the redirect mode](https://github.com/github/fetch/issues/137)
- Inability to [change the cache directive](https://github.com/github/fetch/issues/438#issuecomment-261272466)
- Inability to [disable same-origin cookies](https://github.com/github/fetch/pull/56#issuecomment-68835992)

2. Abort
- `xhr.abort()` first set state to done, finally response to a
[network error](https://xhr.spec.whatwg.org/#the-abort()-method);
- [fetch](https://dom.spec.whatwg.org/#aborting-ongoing-activities) will reject promise with a new "AbortError" DOMException.

As implemented in `fake_xml_http_request`, the request is resolved once its
state is changed to `DONE`.
So the scenario happens in pretender is:
  1). state changes to `DONE`, trigger resolve request
  2). abort, trigger reject
  3). xhr.onerror, trigger reject
The first resolve wins, error thus not rejected but an empty request is resolved.

3. Though polyfilled by xhr, fetch returns a Promise and is asynchronous by
nature.

4. No streaming https://github.com/github/fetch/issues/198


xg-wang:
Yetch doesn't provide a ponyfill now, include the file temporarily.

xg-wang:
This PR should wait for the release of https://github.com/github/fetch
Then CI should pass.

billymoon:
Any updates on this PR? I am super keen to consume fetch-pretender from a cdn.

xg-wang:
CI error: 
`bower EINVRES       Request to https://bower.herokuapp.com/packages/FakeXMLHttpRequest failed with 502`
likely to be fixed by restarting

xg-wang:
This is ready now. Need review @stefanpenner @rwjblue

test/fetch_test.js
```diff
@@ -0,0 +1,123 @@
...
+
+    this.pretender.get(
+      '/some/path',
+      function(request) {
+        return [200, {}, ''];
+      },
+      false
+    );
+
+    fetch('/some/path').then(function() {
```
stefanpenner: This is async, so this needs to block

test/fetch_test.js
```diff
@@ -0,0 +1,123 @@
...
+      },
+      200
+    );
+
+    var controller = new AbortController();
+    var signal = controller.signal;
+    setTimeout(function() {
+      controller.abort();
+    }, 10);
+    fetch('/downloads', { signal: signal })
```
stefanpenner: if this is qunit lets use the async promise assert helpers

test/fetch_test.js
stefanpenner: if this is qunit lets use the async promise assert helpers

stefanpenner:
Oops, I merged this prematurely. @xg-wang can you address the above two comments today? If not, i'll have to revert until addressed.

PR #237, opened 2018-11-07
Export Pretender in a way Typescript understands

### LINKED ISSUE #229: Import Pretender in Typescript yields empty module
Hi!

I'm trying to use Pretender in a Glimmer application but I'm having trouble importing the module. Using `import Pretender from 'pretender'` doesn't resolve anything because there's apparently no default export. I'm now using `import * as pretender from 'pretender'` to get the whole thing but I just get an empty object. 

I saw there's type support and tried to read the related issues (that are now closed) but I still don't know how to make it work properly.

Am I missing something? 

Thanks,
Jorge L

happycollision:
Fixes #229 

- [x] Test in TS environment
- [x] run test suite to ensure the rest is working
- [ ] get feedback from people who know commonjs to find out if this breaks other things (please comment below)

This patch fixes my issue, so I am happy with it. I do not know if changing `module.exports` to `module.exports.default` has consequences for others.

### DIFF
--- a/pretender.js
+++ b/pretender.js
@@ -497,7 +497,7 @@ Pretender.Hosts = Hosts;
 Pretender.Registry = Registry;
 
 if (typeof module === 'object') {
-  module.exports = Pretender;
+  module.exports.default = Pretender;
 } else if (typeof define !== 'undefined') {
   define('pretender', [], function() {
     return Pretender;


happycollision:
Seems like TS changed its import behavior to support the original way. Sorry for anyone still experiencing problems related to #229, but it seems that changing the code here is not the solution.

PR #238, opened 2018-11-08
Expose a way to passthrough after the request is found to be unhandled

happycollision:
This PR aims to allow a request to be passed-through in the `unhandledRequests` hook.

Totally fine to say that you don't want this feature. I am adding it for me, but since it was simple, I figured add PR you here.

This allows you to actually pass certain unhandled requests through at runtime if you so choose. Sam Selikoff offered this as an example of a feature this unlocks (better than my original example). Currently, you can't really simply declare multiple multiple domains as a passthrough. You'd need something like:

```js
this.get('api1.something.com', this.passthrough)
this.get('api2.something.com', this.passthrough)
// etc
```

This feature addition allows you to handle all of them with any runtime check you'd like to implement:

```js
server.unhandledRequest = function(verb, path, request) {
  if (path.startsWith('aws')) {
    request.passthrough() // <-- This is what is now possible
  }
}
```

- [x] Add a `.passthrough()` method to the FakeRequest
- [x] Add tests.
- [x] Add documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,35 @@ In some case, you will need to force pretender to passthough, just start your se
 const server = new Pretender({ forcePassthrough: true })
 ```
 
+Other times, you may want to decide whether or not to passthrough when the call is made. In that
+case you can use the `.passthrough()` function on the fake request itself. (The [`unhandledRequest`
+property is discussed below](#unhandled-requests).)
+
+```javascript
+server.unhandledRequest = function(verb, path, request) {
+  if (myIgnoreRequestChecker(path)) {
+    console.warn(`Ignoring request) ${verb.toUpperCase()} : ${path}`);
+  } else {
+    console.warn(
+      `Unhandled ${verb.toUpperCase()} : ${path} >> Passing along. See eventual response below.`
+    )
+  
+    const xhr = request.passthrough(); // <-- A native, sent xhr is returned
+  
+    xhr.onloadend = (ev) => {
+        console.warn(`Response for ${path}`, {
+          verb,
+          path,
+          request,
+          responseEvent: ev,
+        })
+      };
+  }
+};
+```
+
+The `.passthrough()` function will immediately create, send, and return a native `XMLHttpRequest`.
+
 ### Timing Parameter
 The timing parameter is used to control when a request responds. By default, a request responds
 asynchronously on the next frame of the browser's event loop. A request can also be configured to respond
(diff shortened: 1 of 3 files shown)


happycollision:
I created this PR a long time ago with the full intent to finish off that checklist above. However, I understand that new features like this may not be desirable in your project, and now that I am back to working on a project that depends on my fork, I'd love to finish this off if it is something that belongs in Pretender. I'd be happy to finish this PR (and also add proper TS typings for the new feature) if someone who maintains this project would let me know if the functionality is desired. I found it immensely helpful to be able to call `.passthrough()` on a request as illustrated in my example code above.

Any thoughts, @bantic, @mike-north, @trek, or @xg-wang?

happycollision:
@xg-wang I went ahead and did this. Tests pass. I hope this feature is something you agree with. I have found it immensely useful.

samselikoff:
Being able to decide whether to passthrough a request at runtime is super useful & actually something we added to Mirage.  I'm in favor!

README.md
```diff
@@ -174,6 +174,27 @@ In some case, you will need to force pretender to passthough, just start your se
...
+
+```javascript
+server.unhandledRequest = function(verb, path, request) {
+  console.warn(
+    `Unhandled ${verb.toUpperCase()} : ${path} >> Passing along. See eventual response below.`
+  )
+
+const xhr = request.passthrough() // <-- The real xhr is returned
+
+xhr.onloadend = (ev) => {
```
xg-wang: Should these 2 lines add an indentation?

test/passthrough_test.js
```diff
@@ -339,4 +339,21 @@ describe('passthrough requests', function(config) {
 
     xhr.send();
   });
+
+  it('allows a passthrough on an unhandledRequest via request.passthrough()', function(assert) {
```
xg-wang: Can you add another test for return type of `request.passthrough()` is `originalXMLHttpRequest` isntance?

happycollision:
I am happy to make the requested changes. I don’t blame you for not commenting sooner. My original example left a little something to be desired (like how is it useful?!).

Thanks for having a look. I’ll get around to those changes either tonight or within a few days.

test/passthrough_test.js
happycollision: Done. I actually should have looked back to the change I made all those months ago before I edited the Readme. It is the same thing that is returned from `createPassthrough`, which is actually a new instance of the native `XMLHttpRequest`. The new test and updated Readme accurately reflect that reality.

happycollision:
I should note that in addition to the adding a test, I altered the readme to more accurately reflect what happens when the user calls the new method.

PR #235, opened 2018-09-14
Switch back to whatwg-fetch

xg-wang:
https://github.com/github/fetch/compare/141665b...master
2 changes worth notice due to fetch spec change:
https://github.com/github/fetch/pull/651 https://github.com/github/fetch/pull/640

PR #223, opened 2017-12-16
TypeScript type definitions

mike-north:
Haven't established whether this is complete yet, but these types are working well in a few apps Pretender-heavy test suites (about 130 acceptance test modules total)

Depends on https://github.com/pretenderjs/FakeXMLHttpRequest/pull/35 being merged, release cut, etc...

Early adopters, please feel free to copy/paste these type definitions into a `pretender.d.ts` file within your project

###### TODO
- [X] https://github.com/pretenderjs/FakeXMLHttpRequest/pull/35 merged
- [x] Major FakeXMLHttpRequest release cut

Issue #221, opened 2017-12-01
npm install and npm usage documentation missing


QuantumInformation:

jherdman:
Alternatively, whatwg-fetch is missing as an import here:

https://github.com/pretenderjs/pretender/blob/6cdf4e5c2383e43de6c9b17d694dd5bee817f6ca/index.d.ts#L1-L2

xg-wang:
@jherdman , if you are on browser environment the assumption is that `route-recognizer`, `fake-xml-http-request` and `whatwg-fetch/dist/fetch.umd.js` are installed together with `pretender`, so they are dependencies. You can take a look at [`ember-cli-pretender`](https://github.com/rwjblue/ember-cli-pretender/blob/master/index.js#L18)'s implementation.

For `index.d.ts`, can you explain more on what should be imported here?

jherdman:
Thank you, that definitely helps. I think we can use this information to beef up the README with installation instructions — things currently are a bit murky.

I'm a bit of a TypeScript neophyte, so I don't fully understand the significance of the `index.d.ts` file. But it seems strange to me that

https://github.com/pretenderjs/pretender/blob/6cdf4e5c2383e43de6c9b17d694dd5bee817f6ca/pretender.js#L19

references code that isn't necessarily loaded. And that

https://github.com/pretenderjs/pretender/blob/6cdf4e5c2383e43de6c9b17d694dd5bee817f6ca/pretender.js#L13-L15

references code that looks to be loaded in the `index.d.ts` file. This line of thinking comes from my above noted `FakeFetch` error, but I don't see a RouteRecognizer error (maybe it's loaded as a side dependency already???).

xg-wang:
@jherdman you don't need to worry about `index.d.ts` file since mirage is not using typescript. If you're curious about how typescript def loading works, you can checkout their [documentation](https://www.typescriptlang.org/docs/handbook/declaration-files/consumption.html).

jherdman:
Thank you again for the reply, @xg-wang . I really appreciate your time.

PR #243, opened 2019-01-02
Add build step and TS support

givanse:
Depends on:
 - [x] https://github.com/pretenderjs/pretender/pull/241
 - [x] https://github.com/pretenderjs/FakeXMLHttpRequest/pull/46


README.md
```diff
@@ -40,6 +38,23 @@ const server = new Pretender(function() {
...
+```
+
+Or as a module.
+
+```javascript
+import Pretender from 'pretender';
+const server = new Pretender(function() {});
+```
+
+Full example: [use-pretender-as-a-module](https://github.com/givanse/use-pretender-as-a-module)
```
givanse: For the repo linked here, I added as collaborators the people that have merge access to Pretender.

Or if you prefer me to delete the mention of 'use-pretender-as-a-module', that is fine, I will.

package.json
```diff
@@ -1,20 +1,23 @@
...
-  "main": "./pretender.js",
+  "main": "./dist/pretender.js",
+  "module": "./src/pretender.es.js",
   "description": "Pretender is a mock server library for XMLHttpRequest and Fetch, that comes with an express/sinatra style syntax for defining routes and their handlers.",
   "license": "MIT",
   "engines": {
     "node": ">=4"
   },
   "scripts": {
+    "prepublishOnly": "npm run build && npm run tests-only",
```
givanse: Making sure that the build step is ran before publishing a new version, in case a PR bypassed a red CI somehow.

package.json
```diff
@@ -1,20 +1,23 @@
...
     "pretest": "bower install",
+    "build": "rollup --config",
     "test": "npm run lint && npm run jscs && npm run tests-only",
-    "test-ci": "npm run pretest && npm run lint && npm run jscs && npm run tests-only-ci",
+    "test-ci": "npm run pretest && npm run build && npm run lint && npm run jscs && npm run tests-only-ci",
```
givanse: Making sure that the build step is ran before running the tests, in case there is a PR that didn't include the build outputs.

package.json
```diff
@@ -36,12 +39,19 @@
...
     "phantomjs": "^2.1.7",
     "qunit": "^2.6.1",
-    "sinon": "^3.2.1"
+    "rollup": "0.68.2",
+    "rollup-plugin-commonjs": "^9.2.0",
+    "rollup-plugin-multi-entry": "^2.1.0",
+    "rollup-plugin-node-resolve": "^4.0.0",
+    "rollup-plugin-typescript": "^1.0.0",
+    "sinon": "^3.2.1",
+    "tslib": "^1.9.3",
```
mike-north: where is tslib being used?

README.md
```diff
@@ -12,8 +12,6 @@ with an express/sinatra style syntax for defining routes and their handlers.
 Pretender will temporarily replace native XMLHttpRequest and Fetch , intercept
 all requests, and direct them to little pretend service you've defined.
 
-**:warning: Pretender only works in the browser!**
-
```
mike-north: Removing this would seem to indicate that pretender works in non-browser environments now. Is this the case?

README.md
mike-north: What purpose does the example serve, other than proving that the above

```ts
import Pretender from 'pretender';
const server = new Pretender(function() {});
```
is possible?

src/index.ts
```diff
@@ -144,7 +131,7 @@ function Pretender(/* routeMap1, routeMap2, ..., options*/) {
   self.XMLHttpRequest = interceptor(ctx);
 
   // polyfill fetch when xhr is ready
-  this._fetchProps = FakeFetch ? ['fetch', 'Headers', 'Request', 'Response'] : [];
```
mike-north: This was just added in a recently merged PR #235, so I'm surprised to see you changing it back. Have you studied the other PR, and verified that your change is correct?

tsconfig.json
```diff
@@ -0,0 +1,60 @@
...
+    // "preserveSymlinks": true,              /* Do not resolve the real path of symlinks. */
+
+    /* Source Map Options */
+    // "sourceRoot": "",                      /* Specify the location where debugger should locate TypeScript files instead of source locations. */
+    // "mapRoot": "",                         /* Specify the location where debugger should locate map files instead of generated locations. */
+    // "inlineSourceMap": true,               /* Emit a single file with source maps instead of having a separate file. */
+    // "inlineSources": true,                 /* Emit the source alongside the sourcemaps within a single file; requires '--inlineSourceMap' or '--sourceMap' to be set. */
+
+    /* Experimental Options */
+    "experimentalDecorators": true,           /* Enables experimental support for ES7 decorators. */
```
mike-north: I don't see any decorator use. Does this option need to be enabled?

tsconfig.json
```diff
@@ -0,0 +1,60 @@
+{
+  "compilerOptions": {
+    /* Basic Options */
+    "target": "es5",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */
+    "module": "commonjs",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
+    // "lib": [],                             /* Specify library files to be included in the compilation. */
+    "allowJs": true,                         /* Allow javascript files to be compiled. */
```
mike-north: Are we compiling JS via `tsc`? Does this option need to be enabled?

src/pretender.es.js
```diff
@@ -0,0 +1,950 @@
+import self$1 from './iife-self-placeholder.js';
```
mike-north: Is this a build output? Should it be checked into the repo?

givanse:
@mike-north thanks for the feedback! I'll be addressing each comment.

package.json
givanse: `tslib` is a dependency used by [rollup-plugin-typescript](https://github.com/rollup/rollup-plugin-typescript)

>Note that both typescript and tslib are peer dependencies of this plugin that need to be installed separately.

tsconfig.json
givanse: Right, there aren't any, so I'll turn it off.

tsconfig.json
givanse: For the objective of this PR it isn't needed, so turning it off.

Strictly speaking, I don't think its using `tsc` (still very new to this ecosystem). `rollup-plugin-typescript` grabs and uses the API from `typescript/lib/typescript.js`. And looks like `tsc` is purely the CLI tool to use the same APIs.

src/index.ts
givanse: My initial impression was that it was just a fallback put in place to avoid an error state. (had seem to be a reaction to the issue linked to in the original PR)

Now I see that I was wrong, it is about restoring the global space to its original state **if** needed. I would have used different syntax, but that is out of scope for this PR :)

givanse: The thing is, I still don't see how would `FakeFetch` ever be undefined? Originally, the conversation around it started because of issue #221, but that one is easily explained. The html wasn't loading `fetch.umd.js`.

I don't think we need the fallback, but I'm going to leave that in place since it isn't immediately pertinent to the objective of this PR.

tsconfig.json
mike-north: Ah, I didn't mean `tsc` directly. As of a couple of weeks ago, there are two options for transforming TS code to JS code: the typescript compiler (which actually compiles, and emits JS), and babel 7 (which just knows how to eat the non-javascript parts, leaving only JS, but no type-checking whatsoever).

In this rollup plugin the contents of `tsconfig.json` are read

https://github.com/rollup/rollup-plugin-typescript/blob/7f553800b90fc9abdcd13636b17b4824459d3960/src/options.js#L40-L52

which is then passed to the typescript compiler

https://github.com/rollup/rollup-plugin-typescript/blob/7f553800b90fc9abdcd13636b17b4824459d3960/src/index.js#L92-L96

so whatever we do here is _as if_ we were just running `tsc` on the command line.

src/index.ts
mike-north: @xg-wang can you comment here as to the correct behavior? I don't want anyone to roll back a legitimate bugfix

README.md
givanse: It is a full implementation showing how to consume each build output. But yeah, it is out of scope for the README because that is only one recipe out of many. Maybe I'll turn it into a blog spot or something like that.

README.md
givanse: Nope, I had the wrong idea about this when I started to play with the Rollup config. Reverting, thanks!

README.md
mike-north: Let's omit it. Your description of how to consume pretender as an ES module is sufficient here.

rollup.config.js
```diff
@@ -0,0 +1,43 @@
...
+      name: 'Pretender',
+      file: pkg.main,
+      format: 'iife',
+      globals: {
+        [selfId]: 'self',
+        'whatwg-fetch': 'FakeFetch',
+        'fake-xml-http-request': 'FakeXMLHttpRequest',
+        'route-recognizer': 'RouteRecognizer',
+      },
+      banner: '//TODO: remove from source control when the tests begin using the source file instead of this one.\n' +
```
givanse: I want to raise visibility on this code comment. The future direction I see Pretender taking is to run the tests by loading Pretender as a module (src/index.ts) instead of the iife.

src/pretender.es.js
givanse: Nope, build output removed.

dist/pretender.js
```diff
@@ -0,0 +1,435 @@
+//TODO: remove from source control when the tests begin using the source file instead of this one.
```
mike-north: should this file committed?

givanse: I thought we did, because it replaces `./pretender.js`, but we don't. `test-ci` triggers a build, duh!

mike-north:
We have to think about how to deal with type information in light of this PR. Typically, after TS conversion is complete, ambient declaration files (`*.d.ts`) are not necessary anymore (and are in fact a pain to maintain, since they can drift away from the _real_ types in `.ts` code)

Ultimately (i.e., before we publish a new release) we want type information to come from the `.ts` source (via a generated `.d.ts` file) and _not_ from my handwritten definitions (#223). Your TS configuration here is not strict enough to provide real type safety, so here's what I suggest

- [x] Leave the `index.d.ts` in place
- [ ] Ensure that rollup generates no declaration files
- [ ] Try your branch in a real project, and ensure that type information (via `index.d.ts` is available to consumers)

src/index.ts
xg-wang: When `appearsBrowserified` is `false`, if the `whatwg-fetch` lib is not installed `self.WHATWGFetch` will be `undefined`. This always happens for non `ember-cli-pretender` users since they have to manual bundle the fetch lib.

In this pr the build output will be:
```js
var FakeFetch = self.WHATWGFetch;
var Pretender = (function (self, RouteRecognizer, FakeXMLHttpRequest, FakeFetch) {
}(self, RouteRecognizer, FakeXMLHttpRequest, FakeFetch));
```

The `self.WHATWGFetch` is pulled in by https://github.com/pretenderjs/pretender/blob/v2.1.1/karma.conf.js#L24

src/index.ts
```diff
@@ -1,21 +1,7 @@
-(function(self) {
-'use strict';
-
-function getModuleDefault(module) {
-  return module.default || module;
-}
-
-var appearsBrowserified = typeof self !== 'undefined' &&
```
xg-wang: Is this a breaking change?

givanse: It shouldn't break anyones app. The build step produces two files with unique consumption mechanisms that immediately inform whether you are *browserifying* or not depending on which one you choose to use. So, Pretender doesn't need to figure that out at runtime anymore.

givanse:
I linked it to ember-cli-mirage and ran all the tests (`npm run test:all`), all passed.

Adding the types blew out the diffs. If you prefer, I'm okay to go back to plan B (not generate the declaration file) and postpone the addition of the types for a subsequent PR.

mike-north:
@givanse Yes, let's ensure rollup does not generate declaration files, and stick with the handwritten types until we can get the `src/index.ts` code up to a level of type safety such that the emitted types will be suitable for consumer use

givanse:
Yeah, it is ready. Thanks for your help.

cowboyd:
This is excellent! Can we get a release based on this version? We're using Pretender to in an Angular app and installing off of master fixes the manual module resolution problems that we were encountering 👌

PR #248, opened 2019-02-06
output umd

givanse:
`yarn build` will produce `dist/pretender.umd.js`

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,10 @@
     "route-recognizer": "^0.3.3",
     "whatwg-fetch": "^3.0.0"
   },
+  "files": [
+    "dist",
+    "index.d.ts"
+  ],
   "jspm": {
     "shim": {
       "pretender": {
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "pretender",
   "version": "2.1.0",
   "main": "./dist/pretender.js",
-  "module": "./src/pretender.es.js",
+  "module": "./dist/pretender.es.js",
   "types": "index.d.ts",
   "description": "Pretender is a mock server library for XMLHttpRequest and Fetch, that comes with an express/sinatra style syntax for defining routes and their handlers.",
   "license": "MIT",
@@ -40,7 +40,7 @@
     "karma-sinon": "^1.0.5",
     "phantomjs": "^2.1.7",
     "qunit": "^2.6.1",
-    "rollup": "0.68.2",
+    "rollup": "^1.1.2",
     "rollup-plugin-commonjs": "^9.2.0",
     "rollup-plugin-multi-entry": "^2.1.0",
     "rollup-plugin-node-resolve": "^4.0.0",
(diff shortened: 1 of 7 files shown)


mike-north:
have you verified that this works? When I tried something similar to upgrade ember-cli-pretender, it seemed like pretender was looking for `iife-self-placeholder.js` as an independent module

mike-north:
screen shot 2019-02-06 at 2 46 32 pm
seems like this is still a problem

givanse:
Yeah, that doesn't look right. The only verification I did was to load it in a regular HTML file and follow the invocation of the UMD module, which seemed fine, but I didn't catch that. I guess I'll try to make the ember-cli-pretender upgrade to be sure.

givanse:
Also, I can add a couple of tests for the built files if we get this merged https://github.com/pretenderjs/pretender/pull/244

givanse:
I've updated the config, the UMD output now has the correct AMD invocation.

```
typeof define === 'function' && define.amd ?
    define('pretender', ['route-recognizer', 'fake-xml-http-request', 'whatwg-fetch'], factory)
```

The reasons why using this new UMD file in ember-cli-pretender doesn't work are:

a) [The AMD spec requires an `amd` property declared for the `define` function.](https://github.com/amdjs/amdjs-api/blob/master/AMD.md#defineamd-property-) and the `define` function in the loader that the addon is using doesn't have it. (ember-cli/loader.js) This isn't an oversight of the loader and there might be ways to solve for it (https://github.com/ember-cli/loader.js/commit/ce27cff63c10380559b49ef9979ca80619bc5fbb).

b) ember-cli-pretender is relying on how Pretender was making use of the `define` function to just re-export a global without really using AMD modules, like this:
```
define('pretender', [], function() {
  return Pretender;
});
```

I question the need to support an upgrade path for ember-cli-pretender since what its meant to do is allow you to:
 - import Pretender into a project using ES6 syntax
 - save you the trouble of having to think about which and how to import the dependencies it needs

Considering that now this library provides an ES6 module that provides both of those things ([for example](https://github.com/givanse/glimmer-app-w-pretender/blob/master/src/ui/components/GlimmerAppImportNpmEs6/component-test.ts#L3)), adding the hacks that it'll take to make the UMD module with ember-cli-pretender doesn't seem worth it.

givanse:
ember-cli-pretender could continue to use the file with the IIFE if we kept using the `define` just to re-export the Pretender global, as before.

And, in any case, should Pretender still make available a UMD file?

package.json
```diff
@@ -54,6 +54,10 @@
     "route-recognizer": "^0.3.3",
     "whatwg-fetch": "^3.0.0"
   },
+  "files": [
+    "dist",
+    "index.d.ts"
+  ],
```
givanse: This is way better than `.npmignore`.

givanse:
ember-cli-using the IIFE and then handling from within the re-export of the global: https://github.com/rwjblue/ember-cli-pretender/pull/70

givanse:
I'll create a new PR for this.

PR #256, opened 2019-04-12
refactor: extract hosts

givanse:
- no logic changes
 - extract `Hosts` to its own class
 - updated a few `var`s to `let`s

### DIFF
--- a/src/hosts.ts
+++ b/src/hosts.ts
@@ -1,0 +1,33 @@
+import Registry from './registry';
+import parseURL from './parse-url';
+
+/**
+ * Hosts
+ *
+ * a map of hosts to Registries, ultimately allowing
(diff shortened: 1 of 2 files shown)


src/hosts.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Hosts
+ *
+ * a map of hosts to Registries, ultimately allowing
+ * a per-host-and-port, per HTTP verb lookup of RouteRecognizers
+ */
+export default class Hosts {
+
+  private registries = {};
```
xg-wang: Build outputs would be `Hosts.registries` rather than `Hosts._registries` before. Not sure whether it should be considered as breaking change.

PR #267, opened 2019-09-04
use `url-parse` package as parsing fallback

FLGMwt:
Same as #219, hoping to get pretender working with React Native projects that don't have `document.createElement` available for URL parsing. `url-parse` has a minimal minimal footprint and this won't affect other consumers. That said, I'm happy to rework it to not use `document.createElement` at all and rely only on `url-parse` for consistency.

As for the `self` reference that #219 offered, I've found that it's common for RN projects using browser-y libraries to use [react-native-browser-polyfill](https://github.com/johanneslumpe/react-native-browser-polyfill) to achieve something similar, so changes aren't necessary here. Otherwise, I can pull in the [suggestion](https://github.com/pretenderjs/pretender/pull/219#discussion_r151718023) provided in #219 if that's preferred. 

Thanks : D

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
   "dependencies": {
     "fake-xml-http-request": "^2.0.0",
     "route-recognizer": "^0.3.3",
+    "url-parse": "^1.4.7",
     "whatwg-fetch": "^3.0.0"
   },
   "files": [
@@ -65,7 +66,8 @@
         "deps": [
           "route-recognizer",
           "fake-xml-http-request",
-          "whatwg-fetch"
+          "whatwg-fetch",
+          "url-parse"
         ],
         "exports": "Pretender"
       }
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -6,15 +6,16 @@ const fs = require('fs');
 const globals = {
   'whatwg-fetch': 'FakeFetch',
   'fake-xml-http-request': 'FakeXMLHttpRequest',
-  'route-recognizer': 'RouteRecognizer'
+  'route-recognizer': 'RouteRecognizer',
+  'url-parse': 'ParsedUrl'
 };
 
 const rollupTemplate = fs.readFileSync('./iife-wrapper.js').toString();
 const [ banner, footer ] = rollupTemplate.split('/*==ROLLUP_CONTENT==*/');
-
 module.exports = {
   input: 'src/index.ts',
-  external: Object.keys(pkg.dependencies),
+  // don't exclude url-parse because it does *not* provide a UMD bundle and needs to be included
+  external: Object.keys(pkg.dependencies).filter(x => x !== 'url-parse'),
   output: [
     {
       name: 'Pretender',
--- a/src/parse-url.ts
+++ b/src/parse-url.ts
@@ -1,3 +1,5 @@
+import { default as ParsedUrl } from 'url-parse';
+
 /**
  * parseURL - decompose a URL into its parts
  * @param  {String} url a URL
@@ -16,32 +18,30 @@
  * }
  */
 export default function parseURL(url: string) {
-  // TODO: something for when document isn't present... #yolo
-  var anchor = document.createElement('a');
-  anchor.href = url;
+  var parsedUrl = new ParsedUrl(url);
 
-  if (!anchor.host) {
+  if (!parsedUrl.host) {
     // eslint-disable-next-line no-self-assign
-    anchor.href = anchor.href; // IE: load the host and protocol
+    parsedUrl.href = parsedUrl.href; // IE: load the host and protocol
   }
 
-  var pathname = anchor.pathname;
+  var pathname = parsedUrl.pathname;
   if (pathname.charAt(0) !== '/') {
     pathname = '/' + pathname; // IE: prepend leading slash
   }
 
-  var host = anchor.host;
-  if (anchor.port === '80' || anchor.port === '443') {
-    host = anchor.hostname; // IE: remove default port
+  var host = parsedUrl.host;
+  if (parsedUrl.port === '80' || parsedUrl.port === '443') {
+    host = parsedUrl.hostname; // IE: remove default port
   }
 
   return {
     host: host,
-    protocol: anchor.protocol,
-    search: anchor.search,
-    hash: anchor.hash,
-    href: anchor.href,
+    protocol: parsedUrl.protocol,
+    search: parsedUrl.query,
+    hash: parsedUrl.hash,
+    href: parsedUrl.href,
     pathname: pathname,
-    fullpath: pathname + (anchor.search || '') + (anchor.hash || '')
+    fullpath: pathname + (parsedUrl.query || '') + (parsedUrl.hash || '')
   };
 }
--- a/iife-wrapper.js
+++ b/iife-wrapper.js
@@ -15,6 +15,9 @@ var Pretender = (function(self) {
   var FakeXMLHttpRequest = appearsBrowserified
     ? getModuleDefault(require('fake-xml-http-request'))
     : self.FakeXMLHttpRequest;
+  var ParsedUrl = appearsBrowserified
+    ? getModuleDefault(require('url-parse'))
+    : self.ParsedUrl;
 
   // fetch related ponyfills
   var FakeFetch = appearsBrowserified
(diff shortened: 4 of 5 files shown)


samselikoff:
👍 I'm in favor of this change – and I also think we should drop the `try/catch` and just use `url-parse`! The lib is popular & well-maintained, don't think there's any reason to preserve the original behavior.

FLGMwt:
Mk, updated to use `url-parse` unconditionally and `yarn test` browser-based tests are fine, but `yarn test-ci` in Phantom is sad. 

Any ideas? Am I missing something needed for including the url-parse pakage?

samselikoff:
Ah. Maybe you can just map it and call it something else?

FLGMwt:
> Ah. Maybe you can just map it and call it something else?

Yeah, found another issue after that though w/ the new name not being available to the iife. Couldn't figure out how the other deps were working since I was doing the same thing as them. My best guess is that whatwg and the two bower deps had umd builds that exposed themselves on global but url-parse doesn't. Fix here is to tell rollup *not* to mark it as a `external`/already dealt-with dep so it gets included in the rollup. I think? 🤷‍♂

samselikoff:
LGTM!

I think we don't need to include url-parse in the `es` build but in master all the config is shared. We're doing work on the Rollup config in our PR so I vote we merge this and then we can improve the es build either in our PR or a separate one. Keep it small ya know.

package.json
```diff
@@ -53,6 +53,7 @@
   "dependencies": {
     "fake-xml-http-request": "^2.0.0",
     "route-recognizer": "^0.3.3",
+    "url-parse": "^1.4.7",
```
xg-wang: To bundle `url-parse` in the output file, `url-parse` should lives in `devDependencies`.

rollup.config.js
```diff
@@ -6,15 +6,16 @@ const fs = require('fs');
 const globals = {
   'whatwg-fetch': 'FakeFetch',
   'fake-xml-http-request': 'FakeXMLHttpRequest',
-  'route-recognizer': 'RouteRecognizer'
+  'route-recognizer': 'RouteRecognizer',
+  'url-parse': 'ParsedUrl'
```
xg-wang: The global in output file seems to be `urlParse` rather than `ParsedUrl`.

rollup.config.js
```diff
@@ -6,15 +6,16 @@ const fs = require('fs');
...
 };
 
 const rollupTemplate = fs.readFileSync('./iife-wrapper.js').toString();
 const [ banner, footer ] = rollupTemplate.split('/*==ROLLUP_CONTENT==*/');
-
 module.exports = {
   input: 'src/index.ts',
-  external: Object.keys(pkg.dependencies),
+  // don't exclude url-parse because it does *not* provide a UMD bundle and needs to be included
+  external: Object.keys(pkg.dependencies).filter(x => x !== 'url-parse'),
```
xg-wang: This need to be changed when updating how this package is being bundled.

src/parse-url.ts
```diff
@@ -16,32 +18,30 @@
...
-  // TODO: something for when document isn't present... #yolo
-  var anchor = document.createElement('a');
-  anchor.href = url;
+  var parsedUrl = new ParsedUrl(url);
 
-  if (!anchor.host) {
+  if (!parsedUrl.host) {
     // eslint-disable-next-line no-self-assign
-    anchor.href = anchor.href; // IE: load the host and protocol
+    parsedUrl.href = parsedUrl.href; // IE: load the host and protocol
```
xg-wang: I believe the comment was for IE quirks. This should be changed when moving away from anchor tag method.

src/parse-url.ts
```diff
@@ -16,32 +18,30 @@
...
   if (pathname.charAt(0) !== '/') {
     pathname = '/' + pathname; // IE: prepend leading slash
   }
 
-  var host = anchor.host;
-  if (anchor.port === '80' || anchor.port === '443') {
-    host = anchor.hostname; // IE: remove default port
+  var host = parsedUrl.host;
+  if (parsedUrl.port === '80' || parsedUrl.port === '443') {
+    host = parsedUrl.hostname; // IE: remove default port
```
xg-wang: Same as my comment above. These seem to be addressing IE issue, we should have test coverage on the mentioned cases when moving to `url-parse`.
Although there's no cross browser setup yet, but it's still worth adding test for this.

samselikoff:
> If we are going to include a polyfill for parsing url (probably only for IE), an alternative is to include core-js's URL polyfill

@xg-wang The reason we're bringing in `url-parse` is not to polyfill old browsers but to have better cross-environment support, so Pretender can be used in JS environments that don't have `document`. (In this case, that was React Native, but you could imagine wanting to use Pretender in a non-browser node environment too.)

If consumers need to polyfill ie11 behavior they should be able to use `core-js` on their own (and have it apply to not only Pretender but all their dependencies...)

What do you think?

samselikoff:
And I agree it would be nice to get test coverage for ie11. But not exactly sure how. (I guess ie11 test coverage is more important for the iife build as folks using the es build will have a build step and should be able to use core-js on their own.)

xg-wang:
`URL` constructor is available since node 7 https://nodejs.org/api/url.html#url_the_whatwg_url_api

samselikoff:
Oh I see – you're saying this PR could use [the URL class](https://developer.mozilla.org/en-US/docs/Web/API/URL) instead of `document.createElement` _or_ `url-parse`?

xg-wang:
Oh sorry I'm not familiar with React Native, seems it's not implemented https://github.com/facebook/react-native/issues/16434, and the WIP implementation just throws error https://github.com/facebook/react-native/blob/v0.60.5/Libraries/Blob/URL.js#L111.

I think using `url-parse` is fine here, parsing is the only functionality pretender is using.

samselikoff:
Gotcha 👍 

Ok so where does that leave us? (1) remove the ie11 comments (2) and what should we do about core js/ ie11? Are you wanting to ensure the iife build works in ie11, or that an Ember app does that's using the es build?

xg-wang:
@samselikoff https://github.com/unshiftio/url-parse#url-parse has a good test coverage. I'm thinking about bundling and versioning.

Right now pretender is importing `fake-xml-http-request`, `route-recognizer`, and optional `whatwg-fetch`. Should `url-parse` also be treated as external dependency and let consumer import the package? The 3 existing packages all have UMD and ES module distribution but `url-parse` doesn't.

I think what would be easier to consume pretender is to export 
- UMD module `dist/pretender.js` with bundled required packages, that is without `whatwg-fetch`.
- ES module `dist/pretender.es.js` is the ES module version of the UMD module.
- UMD module `dist/pretender.full.umd.js` with all dependencies.

I have a spike at https://github.com/xg-wang/pretender/commit/0c17e6e6c8f575037bf9e669defd35e62fa4ce78

@trek @stefanpenner what's your opinion on bundling? If this approach looks good to you I can create a PR, and @FLGMwt can just address my comments and this PR should be good to go.

samselikoff:
@xg-wang perhaps we can pick this up over in https://github.com/pretenderjs/pretender/pull/269? We started work over there.

Not sure if we need to switch from iife to umd. First question would be, who's using the iife? Supporting all these builds gets complex. I'm leaning more and more these days to focusing on esm build.

The instructions on README for Pretender have said 

> You can load Pretender directly in the browser.
> <script src="pretender.js"></script>

but I'm pretty sure that's wrong, I believe the other dependencies would also need to be loaded, leading me to think nobody is using this, further leading me to think we should not commit to supporting a new UMD build (and definitely not two). It's just more work and the use cases get very complex very fast.

In #269 we preserve the existing iife build and also swap `whatwg-fetch` out for `cross-fetch` during the ESM build, since `whatwg-fetch` cannot be imported in node.

Also looking at your commit, I don't think we want to treat `fake-xml-http-request` or `route-recognizer` as internal to Pretender, we want these to be `import` statements in the final ESM build so that consumers can bring their own and/or their build tools can dedup them if they happen to be using these deps elsehwere.

samselikoff:
@xg-wang want to chat about this in Discord, so we don't duplicate any more work?

samselikoff:
@xg-wang would be great if we get this merged, I don't think we should block @FLGMwt's work on a plan to re-work the rollup config & build system.

What do you think? What work if any is still needed here?

xg-wang:
@samselikoff I think bundling the package is fine, the remaining work should be to address my comments above. @FLGMwt can you take a look?

zomars:
Can I pick up @FLGMwt 's work from here?

xg-wang:
Thank you @FLGMwt I think this is closed by https://github.com/pretenderjs/pretender/pull/288

Issue #268, opened 2019-09-10
Make Pretender importable in node


samselikoff:
👋 Hey all!

As we're getting Mirage to work in different environments, Pretender not being importable is causing us some headaches.

I know that having Pretender _work_ in node is one thing, and that increases the scope of this project and requires more resources for maintenance, but is having it _importable_ something that we could consider committing to?

If so we can take on the work & add some test coverage. That way Mirage can always "just import" Pretender and then decide at run-time whether to instantiate it.

Here's an older issue for reference: https://github.com/pretenderjs/pretender/issues/192

stefanpenner:
@samselikoff sounds good, let's just ensure it is also tested so it does not regress.

samselikoff:
sweet, we're going to work on this today

samselikoff:
@stefanpenner any chance you could shed light on `iife-wrapper`? First time I'm seeing this (though I'm new to rollup & configuring builds). Seems like the `self`/iife there is what's causing the issue.

PR #269, opened 2019-09-13
Ensure Pretender is importable in node

samselikoff:
Fixes https://github.com/pretenderjs/pretender/issues/268


samselikoff:
@stefanpenner think this is ready for review!

LMK if someone else I should @

samselikoff:
This works for Pretender's iife build but not it's ES module build, which we found out when trying to use it in a Gatsby app.

The problem is `whatwg-fetch` is not importable in node.

We're kinda struggling about what to do here. One option would be to vendor `whatwg-fetch` into a local module and wrap it in a `if (typeof window !== undefined)`. That way that local module (and thus the ES build of Pretender) can be imported in any environment.

What do you think? Any tips on how to solve this in another way?

samselikoff:
Think we're in a good spot! Gonna test it on some of my projects now

samselikoff:
Seems to be working everywhere \o/

samselikoff:
There's one more thing...

Need more tests to cover ES build case in non-browser environment.

samselikoff:
To clarify, we need to actually run the real functionality tests against the ES build in a browser-like (`jsdom`) environment. Need to see what `cross-fetch` imports there and ensure it lets Pretender work against `fetch`.

samselikoff:
Closing for now as stale

PR #275, opened 2019-10-21
Replace time-based progress values with byte values. Fixes #274

### LINKED ISSUE #274: Fake progress event contains incorrect "loaded" and "total" values
In my project, I have a unit test for upload functionality - and we use Pretender for mocking the XHR requests.

For the past 3 weeks, we have had a flaky test issue where occasionally a Progress event is triggered with values that are larger than our mock file.

This is essentially what we are doing:

    const mockFile = new File(['a'.repeat(30)], 'text');
    upload(mockFile);
    assert(progressEvent.loaded < mockFile.length);

So we have a mock file that is 30 bytes long, but we were sometimes seeing progressEvent.loaded be well into the hundreds (90s, 100s, 200s, etc) seemingly at random.

I traced this down to a bug at the integration layer between `Pretender` and `fake-xml-http-request`

At [this line](https://github.com/pretenderjs/pretender/blob/master/src/index.ts#L231) you can see that the progress event is being triggered with _time values_ instead of _byte length_ values, which is expected in [fake-xml-http-request](https://github.com/pretenderjs/FakeXMLHttpRequest/blob/master/src/fake-xml-http-request.js#L187) and [ProgressEvent](https://developer.mozilla.org/en-US/docs/Web/API/ProgressEvent). So essentially the `loaded` and `total` fields are being populated with the length of time since the last progress event rather than being informed at all about the file that is being uploaded.

Please fix this! This cost me about 2 weeks to track down. If I get some bandwidth I may submit a pull request.

seanjohnson08:


src/index.ts
```diff
@@ -224,13 +224,17 @@ function verbify(verb) {
...
+function scheduleProgressEvent(request) {
   setTimeout(function() {
     if (!request.aborted && !request.status) {
-      var ellapsedTime = new Date().getTime() - startTime.getTime();
-      request.upload._progress(true, ellapsedTime, totalTime);
-      request._progress(true, ellapsedTime, totalTime);
-      scheduleProgressEvent(request, startTime, totalTime);
+      if (request && request.requestBody && request.requestBody.size) {
+        var size = request.requestBody.size;
+        request.upload._progress(true, size, size);
```
xg-wang: Is this going to always have progress be 100%?

xg-wang: I think the issue here is after `totalTime`, the progress is still being scheduled. Maybe the fix should be stopping scheduling after `totalTime`? At that time the request has been resolved https://github.com/pretenderjs/pretender/blob/master/src/index.ts#L354.

seanjohnson08: Yes, this would always have progress be at 100. I thought about having a progression based on time but decided that faking progress events at 100% would be just as good (certainly better than time diffs) for most use-cases.

As for the scheduling, you are correct that `scheduleProgressEvent` will fire after the `totalTime` (after response sent) but it stops itself because the request will then have a status: https://github.com/pretenderjs/pretender/blob/master/src/index.ts#L229

The current way it's stopping itself is fine and wasn't a problem for my use case.

xg-wang:
Closed by https://github.com/pretenderjs/pretender/pull/276

PR #276, opened 2019-10-24
fix: progress event expects byte size not time

### LINKED ISSUE #274: Fake progress event contains incorrect "loaded" and "total" values

xg-wang:
Spec: 
- https://xhr.spec.whatwg.org/#the-send()-method
- https://xhr.spec.whatwg.org/#concept-event-fire-progress

fix #274 , supersedes #275 
cc @seanjohnson08 @mike-north 

---
TODO:
- [x] all existing tests continue to pass
- [x] we write a new test that would have _failed_ before this code change, and _now passes_ due to the alteration of how the progress events are handled
- [ ] we include great documentation around "handling uploads with Pretender"

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,30 @@ means the route will use the default behavior.
 When the time is right, you can set `externalState` to `"OH NO DOS ATTACK"` which will make all
 future requests take 15 seconds to respond.
 
+#### Scheduling ProgressEvent
+If the timing parameter is resolved as async, then a [`ProgressEvent`](https://xhr.spec.whatwg.org/#interface-progressevent)
+will be scheduled every 50ms until the reqeust has a respond or is aborted.
+
+To listen to the progress, you can define `onprogress` on the `XMLHttpRequest` object or
+its [`upload` attribute](https://xhr.spec.whatwg.org/#the-upload-attribute).
+
+```javascript
+let xhr = new window.XMLHttpRequest();
+xhr.open('POST', '/uploads');
+// https://fetch.spec.whatwg.org/#concept-request-body
+// https://xhr.spec.whatwg.org/#the-send()-method
+let postBody = new ArrayBuffer(8);
+xhr.upload.onprogress = function(event) {
+  // event.lengthComputable === true
+  // event.total === 8
+  // event.loaded will be incremented every ~50ms
+};
+xhr.onprogress = function(event) {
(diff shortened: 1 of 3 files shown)


mike-north:
awesome. I'm fine with treating this as a "feature" and making an accompanying minor version bump, providing
- [ ] all existing tests continue to pass
- [ ] we write a new test that would have _failed_ before this code change, and _now passes_ due to the alteration of how the progress events are handled
- [ ] we include great documentation around this aspect of Pretender, so others don't have to work quite so hard to figure things out as @seanjohnson08 did

xg-wang:
> Looks great! wondering if you can fix the typo in "elapsed"? :)

Good catch! Fixed now.

xg-wang:
@mike-north I pushed 2 commits, first failling test looks like:
```
PhantomJS 2.1.1 (Linux 0.0.0) pretender invoking async requests with `onprogress` upload events in the upload  trigger a progress event each 50ms FAILED
	ProgressEvent has total of requestBody byte size
	Expected: 9
	Actual: 300
```

Also added a section in readme.

PR #278, opened 2019-10-25
Add missing type definitions in index.d.ts

### LINKED ISSUE #277: Incomplete types in index.d.ts
Following #233 (which itself seems solved already) I found a bunch of other (most likely not all) missing type definitions in index.d.ts.

ohcibi:
This adds some of the missing type definitions in index.d.ts. I tried creating the declaration
using tsc but apparently another rollup-typescript plugin is necessary for this to work and I wasn't
able to generate a useful declarations file with it, so I added it manually. Fixes #277

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -20,6 +20,16 @@ export class Server {
   public head: RequestHandler;
 
   public shutdown(): void;
+
+  public map(maps: Function): void;
+
+  public handledRequest(verb: string, path: string, request: FakeXMLHttpRequest & ExtraRequestData): void;
+  public undhandledRequest(verb: string, path: string, request: FakeXMLHttpRequest & ExtraRequestData): void;
+  public passtroughRequest(verb: string, path: string, request: FakeXMLHttpRequest & ExtraRequestData): void;
+  public erroredRequest(verb: string, path: string, request: FakeXMLHttpRequest & ExtraRequestData, error: Error): void;
+
+  public prepareBody(body: string): string;
+  public prepareHeaders(headers: {[k: string]: string}): {[k: string]: string};
 }
 
 export type RequestHandler = (


index.d.ts
```diff
@@ -20,6 +20,16 @@ export class Server {
   public head: RequestHandler;
 
   public shutdown(): void;
+
+  public map(maps: Function): void;
+
+  public handledRequest(verb: string, path: string, request: FakeXMLHttpRequest & ExtraRequestData): void;
+  public undhandledRequest(verb: string, path: string, request: FakeXMLHttpRequest & ExtraRequestData): void;
+  public passtroughRequest(verb: string, path: string, request: FakeXMLHttpRequest & ExtraRequestData): void;
+  public erroredRequest(verb: string, path: string, request: FakeXMLHttpRequest & ExtraRequestData): void;
```
xg-wang:
```suggestion
  public erroredRequest(verb: string, path: string, request: FakeXMLHttpRequest & ExtraRequestData, error: Error): void;
```

index.d.ts
```diff
@@ -20,6 +20,16 @@ export class Server {
...
+
+  public prepareBody(body: string): void;
+  public prepareHeaders(headers: object): void;
```
xg-wang:
```suggestion
  public prepareHeaders(headers: {[k: string]: string}): {[k: string]: string};
```

index.d.ts
```diff
@@ -20,6 +20,16 @@ export class Server {
...
   public shutdown(): void;
+
+  public map(maps: Function): void;
+
+  public handledRequest(verb: string, path: string, request: FakeXMLHttpRequest & ExtraRequestData): void;
+  public undhandledRequest(verb: string, path: string, request: FakeXMLHttpRequest & ExtraRequestData): void;
+  public passtroughRequest(verb: string, path: string, request: FakeXMLHttpRequest & ExtraRequestData): void;
+  public erroredRequest(verb: string, path: string, request: FakeXMLHttpRequest & ExtraRequestData): void;
+
+  public prepareBody(body: string): void;
```
xg-wang:
```suggestion
  public prepareBody(body: string): string;
```

index.d.ts
```diff
@@ -32,6 +42,7 @@ export type ResponseData = [number, { [k: string]: string }, string];
 interface ExtraRequestData {
   params: Params;
   queryParams: QueryParams;
+  requestBody: string
```
xg-wang: https://github.com/pretenderjs/pretender/blob/v3.0.2/src/index.ts#L396
Pretender only adds `params` and `queryParams`. I think this needs to be fixed in `fake-xml-http-request` package.

xg-wang:
Thanks! I think this PR covers all the remaining public APIs mentioned in the readme.

index.d.ts
ohcibi: https://github.com/pretenderjs/FakeXMLHttpRequest/pull/48

PR #279, opened 2019-11-04
fix: remove node engine constraint

xg-wang:
Pretender does not support node at all.

Context: https://github.com/rwjblue/ember-cli-pretender/pull/70#issuecomment-549540587


xg-wang:
This also means Pretender v3 has no breaking change.

PR #280, opened 2019-11-08
support FakeXMLHttpRequest response property

kellyselden:
re https://github.com/pretenderjs/FakeXMLHttpRequest/pull/53

These 2 PRs fixed the problem for me https://github.com/ember-vr/ember-vr-shopping/pull/192

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "typescript-eslint-parser": "^21.0.2"
   },
   "dependencies": {
-    "fake-xml-http-request": "^2.0.0",
+    "fake-xml-http-request": "^2.1.1",
     "route-recognizer": "^0.3.3",
     "whatwg-fetch": "^3.0.0"
   },
--- a/src/index.ts
+++ b/src/index.ts
@@ -113,7 +113,7 @@ function interceptor(ctx) {
     var uploadEvents = [];
 
     // properties to copy from the native xhr to fake xhr
-    var lifecycleProps = ['readyState', 'responseText', 'responseXML', 'responseURL', 'status', 'statusText'];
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -48,7 +48,7 @@
     "typescript-eslint-parser": "^21.0.2"
   },
   "dependencies": {
-    "fake-xml-http-request": "^2.0.0",
+    "fake-xml-http-request": "kellyselden/FakeXMLHttpRequest#response",
```
xg-wang: It seems this package used in test is from bower.
https://github.com/pretenderjs/pretender/blob/v3.0.4/karma.conf.js#L18

xg-wang: Just merged https://github.com/pretenderjs/pretender/pull/281. Can you try rebase?

kellyselden: done

kellyselden:
Do you happen to see where I went wrong in the tests or code?

xg-wang:
@kellyselden It seems the changes in `src/fake-xml-http-request.js` should also be made to `fake-xml-http-request.js` under root, that is the file used for testing https://github.com/pretenderjs/pretender/blob/0b0436294f8d07b56e15cabef58adb55359de6ff/karma.conf.js#L18

kellyselden:
@xg-wang Good call. I did that, and it appears the same issue remains.

package.json
xg-wang: I think it's github artifact cache didn't update. I just tried v2.1.1 and it passed tests.

kellyselden:
Updated with 2.1.1 and all is good now.

PR #288, opened 2020-02-03
Adds url parse and removes the usage of anchor

eluciano11:
## What changed and why
This PR replaces the usage of `document` to parse a URL with the library [`url-parse`](https://github.com/unshiftio/url-parse). By doing this Pretender can be used in environments that don't include the DOM such as React Native.

The work done here is based on this PR https://github.com/pretenderjs/pretender/pull/267. Additionally I added a couple of fixes that @xg-wang requested.

The only review comments that are pending from the old PR are the following:
https://github.com/pretenderjs/pretender/pull/267#discussion_r324433409
https://github.com/pretenderjs/pretender/pull/267#discussion_r324433457

To solve the pending issues I originally thought of using the Karma launcher for IE but that won't work on a IE environment because we won't have an IE binary. I was thinking that we could use SauceLabs or BrowserStack to run the tests on different browsers.

I would appreciate it if @xg-wang or @samselikoff  would share their thoughts on how the IE issue could be solved. If we decide to go with SauceLabs we can request an [Open Source Account](https://saucelabs.com/open-sauce).

### DIFF
--- a/iife-wrapper.js
+++ b/iife-wrapper.js
@@ -15,6 +15,9 @@ var Pretender = (function(self) {
   var FakeXMLHttpRequest = appearsBrowserified
     ? getModuleDefault(require('fake-xml-http-request'))
     : self.FakeXMLHttpRequest;
+  var urlParse = appearsBrowserified
+    ? getModuleDefault(require('url-parse'))
+    : self.urlParse;
 
   // fetch related ponyfills
   var FakeFetch = appearsBrowserified
--- a/package.json
+++ b/package.json
@@ -45,7 +45,8 @@
     "sinon": "^3.2.1",
     "tslib": "^1.9.3",
     "typescript": "~3.1.1",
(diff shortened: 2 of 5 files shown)


samselikoff:
@eluciano11 thanks so much for jumping on this!!

@xg-wang Can we go ahead and merge this, and deal with the IE testing later? Since we don't have IE test coverage for the rest of Pretender I don't think it should be a blocker for this PR.

eluciano11:
@samselikoff no problem, happy to help!

xg-wang:
Hi @eluciano11 is this still draft or it's ready be merged?

eluciano11:
@xg-wang should be good to go!

samselikoff:
Awesome, thanks again!

We can take care of ie coverage elsewhere.

eluciano11:
Np, sounds good!

PR #294, opened 2020-02-20
Upgrade devDependencies and CI

xg-wang:
- Upgrade devDependencies
- Drop phantomjs
- Use yarn and headless Chrome in Travis
- Remove `passthrough request calls jQuery v1 handler` test because we are only including single jQuery version.


xg-wang:
Travis is not happy because phantom is used in ci

xg-wang:
CI passes, can I have a review? @trek @stefanpenner

samselikoff:
I don't know all the details but LGTM + definitely seems an improvement

Issue #293, opened 2020-02-13
Proposal: Vendor whatwg-fetch


samselikoff:
Pretender uses the `whatwg-fetch` package in order to intercept `fetch` calls.

Even though `whatwg-fetch` is explicitly imported [here](https://github.com/pretenderjs/pretender/blob/master/src/index.ts#L2) (and used [here](https://github.com/pretenderjs/pretender/blob/master/src/index.ts#L65)), Pretender is actually relying implicitly on `whatwg-fetch`'s internals to get the `fetch` intercepting behavior working correctly.

The reason is because the `fetch` implementation that `whatwg-fetch` provides actually delegates to `window.XMLHTTPRequest` under the hood. Since Pretender _also_ monkey patches `window.XMLHTTPRequest`, it is able to intercept `fetch` calls equally as well. But it's only due to those internals – due to the fact that the `whatwg-fetch` polyfill happens to use XMLHTTPRequest under the hood. This is not part of `whatwg-fetch`'s public API, even though it is unlikely the change.

The reason this bit us recently is because some folks are trying to use Mirage with Next.js. Next performs some build-time optimizations to ensure apps don't include multiple or heavy versions of popular polyfills, `whatwg-fetch` being one of them. So, they actually swap out the package that `whatwg-fetch` resolves to, meaning when Pretender tries to `import * from 'whatwg-fetch'`, it ends up with a different polyfill entirely, and one that does not happen to use `XMLHTTPRequest` under the hood. Therefore, Pretender (and thus Mirage) don't work.

We asked about how to opt-out of this optimizing behavior [here](https://github.com/zeit/next.js/issues/10509), but as of right now the Zeit team doesn't have any plans to make the behavior configurable.

In any case, regardless of whether Next.js chooses to expose such an option, it is still true that today, Pretender is relying on internals of `whatwg-fetch`, and that poses a risk to Pretender in the event those internals ever change.

I'm proposing that we vendor the latest version of `whatwg-fetch` into Pretender directly, since we are relying on observable but undocumented behavior of its current implementation. This would give us total control over the code, and would also address the Next.js issue since `whatwg-fetch` would no longer be an external dependency that could be hoisted or replaced.

stefanpenner:
It's likely ok to bundle `whatwg-fetch` as part of a dev time build step and presenting them as one bundle to consumers. 

[...]

PR #295, opened 2020-02-27
Vendor whatwg-fetch

samselikoff:
Closes https://github.com/pretenderjs/pretender/issues/293

---

Pretender uses the `whatwg-fetch` package in order to intercept `fetch` calls. However, Pretender relies implicitly on `whatwg-fetch` internals to get the `fetch` intercepting behavior working correctly.

This starting causing some problems in some systems where the `whatwg-fetch` package was swapped or removed altogether, since the package is indeed a polyfill and the system determined the behavior was present and thus the polyfill wasn't needed. However, because Pretender actually relied not only on the polyfill behavior but also internal implementation details, Pretender would break in these systems.

More details: The `fetch` implementation that `whatwg-fetch` provides actually delegates to `window.XMLHTTPRequest` under the hood. Since Pretender _also_ monkey patches `window.XMLHTTPRequest`, it is able to intercept `fetch` calls equally as well. But it's only due to those internals – due to the fact that the `whatwg-fetch` polyfill happens to use XMLHTTPRequest under the hood. This is not part of `whatwg-fetch`'s public API, even though it is unlikely the change.

Since Pretender has been relying on observable but undocumented behavior of `whatwg-fetch`, this commit vendors the package, ensuring that `whatwg-fetch` can no longer be treated as an external dependency and hoisted or replaced.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,5 +1,5 @@
 import FakeXMLHttpRequest from 'fake-xml-http-request';
-import * as FakeFetch from 'whatwg-fetch';
+import * as FakeFetch from './whatwg-fetch';
 import parseURL from './parse-url';
 import Registry from './registry';
 
--- a/iife-wrapper.js
+++ b/iife-wrapper.js
@@ -16,11 +16,6 @@ var Pretender = (function(self) {
     ? getModuleDefault(require('fake-xml-http-request'))
     : self.FakeXMLHttpRequest;
 
-  // fetch related ponyfills
-  var FakeFetch = appearsBrowserified
-    ? getModuleDefault(require('whatwg-fetch'))
-    : self.WHATWGFetch;
-
   /*==ROLLUP_CONTENT==*/
 
   if (typeof module === 'object') {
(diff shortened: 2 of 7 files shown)


samselikoff:
Once we merge this I will make a new PR that will supercede https://github.com/pretenderjs/pretender/pull/269. It will be easy for us now to ensure Pretender is importable in node since we have whatwg-fetch vendored – we can just wrap the module in an `if window` check.

samselikoff:
Closes https://github.com/pretenderjs/pretender/issues/293

samselikoff:
@xg-wang could I get your review on this?

src/index.ts
```diff
@@ -1,5 +1,5 @@
 import FakeXMLHttpRequest from 'fake-xml-http-request';
-import * as FakeFetch from 'whatwg-fetch';
+import * as FakeFetch from './whatwg-fetch';
```
xg-wang: You do not need to copy the whole file, `external: Object.keys(pkg.dependencies),` in rollup config is treating whatwg-fetch as external module, when you remove it from `pkg.dependencies` it will be imported by rollup.
You will still need `whatwg-fetch` in dev dependencies to let rollup find the file.

samselikoff: Yah. But it would make the next step easier since we can wrap it in a check to see if window exists. Otherwise we’ll need to do what I did in the other PR and bring in cross fetch, I think.

xg-wang:
Closed by https://github.com/pretenderjs/pretender/pull/296

PR #298, opened 2020-03-19
Add prettier

ryanto:


xg-wang:
Thanks for adding Prettier! Can you also add a precommit hook to prettify code? https://prettier.io/docs/en/precommit.html

ryanto:
We we're exploring adding this to Mirage, but I got some negative feedback (see https://twitter.com/ryantotweets/status/1203065809167409154). I'm not sure if it's worth adding.

I can add the prettier check to CI so we can failures if someone opens a PR with unformatted code. How does that sound?

xg-wang:
Pretender doesn't run lint in ci now, I think adding prettier eslint integration and make it easier to `yarn eslint --fix` should be enough.

PR #302, opened 2020-05-18
Publish pretender.bundle.js bundling dependencies

xg-wang:
dist/pretender.bundle.js is an iife build with fake-xml-http-request and
route-recognizer bundled to make installation easier.

The default main entry in package.json remains same for compatibility.

Both bundle build and non bundle builds are tested in ci.

### DIFF
--- a/.bowerrc
+++ b/.bowerrc
@@ -1,3 +1,0 @@
-{
-  "directory": "./bower_components"
-}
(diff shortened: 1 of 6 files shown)


xg-wang:
https://github.com/pretenderjs/pretender/releases/tag/v3.4.3

PR #315, opened 2021-04-11
README update

ledleds:
Fixed some typo's and also added an example for a map as when I was implementing this I couldn't work out what it meant by a "map". But I figured it was just like the example I have added. Let me know if i'm wrong though 😄


xg-wang:
Nice, thank you!

PR #320, opened 2021-06-07
Handle progress events for FormData request body properly

### LINKED ISSUE #303: Incorrect "progress" events for requests with FormData
I am trying to build a progress bar and demo / test it with Pretender. This is for a file uploader. I am wrapping the file in a FormData object before uploading: 
```
    const data = new FormData();
    data.append('file', attachedFile.file, attachedFile.file.name);
```
This all works correctly with a real server but not with Pretender. After looking into it a bit I saw that the progress events were sent fine but the event properties are incorrect. There should be two properties in the progress from which one can create a progress bar (or whatever). These are "loaded" and "total". For me they look like this: 

Screenshot 2020-05-18 at 11 11 04

I suspect that this is because Pretender isn't correctly reading the FormData object. In the Pretender code I see this:
```
      if (!body) {
        progressTotal = 0;
      } else {
        // Support Blob, BufferSource, USVString, ArrayBufferView
        progressTotal = body.byteLength || body.size || body.length || 0;
      }
      let progressTransmitted =
        totalTime <= 0 ? 0 : (elapsedTime / totalTime) * progressTotal;
```
https://github.com/pretenderjs/pretender/blob/b38a2b06147f5cb1105eaf7bbc79534b36e25be4/src/index.ts#L63

mrijke:
Calculate the size of a FormData request properly by looping over all the fields
and taking either the size (in case of a File object) or length (in case of a string field).
This fixes #303.

### DIFF
--- a/src/pretender.ts
+++ b/src/pretender.ts
@@ -25,32 +25,36 @@ class NoopArray {
 }
 
 function scheduleProgressEvent(request, startTime, totalTime) {
+  let totalSize = 0;
+  const body = request.requestBody;
+  if (body) {
+    if (body instanceof FormData) {
+      body.forEach((value) => {
+        if (value instanceof File) {
+          totalSize += value.size;
+        } else {
+          totalSize += value.length;
+        }
+      });
+    } else {
+      // Support Blob, BufferSource, USVString, ArrayBufferView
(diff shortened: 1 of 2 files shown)


xg-wang:
Thanks!

PR #322, opened 2021-06-07
Fix #321: improve ResponseHandler types

### LINKED ISSUE #321: Types for RequestHandler are too strict
With the rewrite to TypeScript, the API as described in the README does not work in TypeScript projects.
The following:

```ts
const server = new Pretender(function() {
  this.put('/api/songs/99', request => [404, {}, ""]);
});
```
fails with the following TS error:
```
Argument of type '(request: FakeXMLHttpRequest & ExtraRequestData) => [number, {}, string]' is not assignable to parameter of type 'ResponseHandler'.
  Type '(request: FakeXMLHttpRequest & ExtraRequestData) => [number, {}, string]' is missing the following properties from type 'ResponseHandler': async, numberOfCallsts(2345)
```

The extra attributes that were introduced on the handler instance (`async` and `numberOfCalls`) are also mandatory to provide when defining the mock itself - which I think is not intended, these are meant to be (readonly) properties, correct?

For now the following workaround seems to make TS happy - but it's not very ideal:
```ts
const handler = (): ResponseData => [
  200,
  { "Content-Type": "applicatoin/json" },
  JSON.stringify({ status: "OK!" }),
];
handler.async = false;
handler.numberOfCalls = 0;
```

mrijke:
Fix the ResponseHandler type by introducing a new type that extends
the ResponseHandler and adds the instance-specific attributes to it
(async and numberOfCalls).
This will allow TS clients to use the API as described in the README.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -47,8 +47,11 @@ export type ResponseHandler = {
   (request: FakeXMLHttpRequest & ExtraRequestData):
     | ResponseData
     | PromiseLike<ResponseData>;
+};
+
+export type ResponseHandlerInstance = ResponseHandler & { 
   async: boolean;
   numberOfCalls: number;
-};
+}
 
 export default Server;
--- a/index.d.ts
+++ b/index.d.ts
@@ -36,7 +36,7 @@ export type RequestHandler = (
   urlExpression: string,
   response: ResponseHandler,
   asyncOrDelay?: boolean | number
-) => void;
+) => ResponseHandlerInstance;
 
 export type ResponseData = [number, { [k: string]: string }, string];
 interface ExtraRequestData {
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -47,8 +47,11 @@ export type ResponseHandler = {
   (request: FakeXMLHttpRequest & ExtraRequestData):
     | ResponseData
     | PromiseLike<ResponseData>;
+};
+
+export type ResponseHandlerInstance = ResponseHandler & {
```
xg-wang, line 52: Can we rename `ResponseHandler` to `ResponseHandlerFunc`, and just use `ResponseHandler` for the instances that have `async` and `numberOfCalls`?

xg-wang (changes requested):

index.d.ts
xg-wang: Hmm I take it back, looks like PR #317 actually broke ResponseHandler

xg-wang (approved):

PR #324, opened 2021-06-22
normalize `localhost` to `127.0.0.1`

EndangeredMassa:
When using pretender, I'll sometimes see or write `localhost` and `127.0.0.1` in the same app. It would be handy if these were normalized to the same value.

Here's a stab at it, but I don't mind if there's a better approach. Feel free to close if this isn't worth the trouble.

### DIFF
--- a/src/parse-url.ts
+++ b/src/parse-url.ts
@@ -29,13 +29,18 @@ export default function parseURL(url: string) {
     pathname = '/' + pathname; // IE: prepend leading slash
   }
 
-  var host = parsedUrl.host;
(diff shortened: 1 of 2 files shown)


xg-wang:
Hmmm they do not seem to be exactly the same, at least for the tests being modified. I tried `window.location.host` and they will differentiate the `localhost` vs `127.0.0.1`.

I think a better place is https://github.com/pretenderjs/pretender/blob/ddca21671e3e7c69b0be673eb3cd66007931a379/src/hosts.ts#L24 where we can normalize. What do you think?

PR #331, opened 2021-08-04
Fix setting timeout to sync XHR

xg-wang:
This happens when creating a sync XHR in a normal window environment, regressed from https://github.com/pretenderjs/pretender/pull/326

Failed to set the 'timeout' property on 'XMLHttpRequest': Timeouts cannot be set for synchronous requests made from a document.


elwayman02 (approved):
Thanks for fixing and adding the test to prevent against regressions!

PR #334, opened 2021-08-10
Change test script to first run build

### LINKED ISSUE #333: Implement `prepare` hook to run `yarn build` before `yarn test`
Right now, you have to manually run `yarn build` before running `yarn test`, otherwise the tests fail - it would be helpful to make that automated so that Pretender is always built before tests are run, such that tests can be run successfully locally.

nur557:
Closes #333

### DIFF
--- a/README.md
+++ b/README.md
@@ -497,7 +497,7 @@ server.shutdown(); // all done.
 
 ## Running tests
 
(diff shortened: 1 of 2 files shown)


xg-wang:
Thanks!

Issue #337, opened 2021-10-24
Server class in index.d.ts does not match Pretender class


katehedgpeth:
The Server class defined in [`index.d.ts`](https://github.com/pretenderjs/pretender/blob/master/index.d.ts) does not seem to match the `Pretender` class in [`./src/pretender.ts`](https://github.com/pretenderjs/pretender/blob/master/src/pretender.ts) that it is supposed to represent. It is missing the `handledRequests`, `passthroughRequests`, and `unhandledRequests` properties. There may be more missing types, as I haven't examined every prop closely.

It seems to me that instead of having a separate type definition file that has to stay in sync with the class, it would be better to export the types from the class directly with rollup.

I have a PR ready to go that implements this fix. I'll post it in a few minutes and reference this issue.

AndrewOttavianoAbb:
@katehedgpeth I found the same issue as well! I would like to access `this.handledRequests` for my jest test, but TS is blowing up in my face because the property is not defined on that type. 😢

@xg-wang 
@trek 
@stefanpenner 
@rwjblue

PR #338, opened 2021-10-24
generate index.d.ts in build

katehedgpeth:
I was using `miragejs` which builds off of pretender, and noticed that typescript was complaining that `pretender.handledRequests` does not exist. I discovered that pretender uses a separate `index.d.ts` file that is not referenced by the actual class. I've opened issue #337 to report this, but I'm also submitting a PR with my suggested fix.

It is possible to simply add `class Pretender extends Server`, but that would not throw an error if new properties are added to `Pretender` but not to `Server`. In my opinion, the best thing to do is to export the types directly from the source files in the build.

This PR updates the rollup config to include all types in the build, using `rollup-plugin-typescript2`. Doing so also caught a few typescript errors, which I fixed.

I have created a PR on my fork which proves that this change does not affect the generated `.js` files: https://github.com/katehedgpeth/pretender/pull/1

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "3.4.7",
   "main": "./dist/pretender.js",
   "module": "./dist/pretender.es.js",
-  "types": "index.d.ts",
+  "types": "./dist/index.d.ts",
   "description": "Pretender is a mock server library for XMLHttpRequest and Fetch, that comes with an express/sinatra style syntax for defining routes and their handlers.",
   "license": "MIT",
   "scripts": {
@@ -42,6 +42,7 @@
     "release-it-lerna-changelog": "^2.3.0",
     "rollup": "^2.10.2",
     "rollup-plugin-typescript": "^1.0.0",
+    "rollup-plugin-typescript2": "^0.30.0",
     "sinon": "^9.0.2",
     "tslib": "^1.9.3",
     "typescript": "~3.1.1",
--- a/src/interceptor.ts
+++ b/src/interceptor.ts
@@ -2,7 +2,7 @@ import FakeXMLHttpRequest from 'fake-xml-http-request';
 import { createPassthrough } from './create-passthrough';
 
 export function interceptor(ctx) {
-  function FakeRequest() {
+  function FakeRequest(this: any) {
     // super()
     FakeXMLHttpRequest.call(this);
   }
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,6 +1,6 @@
 import commonjs from "@rollup/plugin-commonjs";
 import resolve from "@rollup/plugin-node-resolve";
-import typescript from "rollup-plugin-typescript";
+import typescript from "rollup-plugin-typescript2";
 import { readFileSync } from "fs";
 import pkg from "./package.json";
 
@@ -31,7 +31,7 @@ module.exports = [
         format: "es",
       },
     ],
-    plugins: [commonjs(), resolve(), typescript()],
+    plugins: [commonjs(), resolve(), typescript({ useTsconfigDeclarationDir: true })],
   },
   {
     input: "src/index.ts",
@@ -42,6 +42,6 @@ module.exports = [
         format: "iife",
       }
     ],
-    plugins: [commonjs(), resolve(), typescript()],
+    plugins: [commonjs(), resolve(), typescript({ useTsconfigDeclarationDir: true })],
   },
 ];
--- a/src/create-passthrough.ts
+++ b/src/create-passthrough.ts
@@ -3,7 +3,7 @@ export function createPassthrough(fakeXHR, nativeXMLHttpRequest) {
   var evts = ['error', 'timeout', 'abort', 'readystatechange'];
 
   // event types to handle on the xhr.upload
-  var uploadEvents = [];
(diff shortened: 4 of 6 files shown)


xg-wang (changes requested):
This is very nice. Thank you for putting this up!

Can you also export the remaining types in https://github.com/pretenderjs/pretender/blob/65fbf0a608f2e4b4fd79657a77a41f52077074a9/index.d.ts? This will ensure type compatibility.

Another request is can you annotate the methods that are in the index.d.ts? For example, `handledRequest` params are currently generated to be typed any, but explicitly typed in index.d.ts. There are couple more there. The concern is there needs another major bump to avoid breaking types if we loosen types in this PR.
- https://github.com/pretenderjs/pretender/blob/65fbf0a608f2e4b4fd79657a77a41f52077074a9/src/pretender.ts#L306
- https://github.com/pretenderjs/pretender/blob/65fbf0a608f2e4b4fd79657a77a41f52077074a9/index.d.ts#L26

Since this repo ignores the generated files, this is new type file I built from this branch:

<details>
<summary>Click to toggle content</summary>

```typescript
import FakeXMLHttpRequest from 'fake-xml-http-request';
import { Params, QueryParams } from 'route-recognizer';
import { ResponseHandler, ResponseHandlerInstance } from '../index.d';
import Hosts from './hosts';
import parseURL from './parse-url';
import Registry from './registry';
interface ExtraRequestData {
    url: string;
    method: string;
    params: Params;
    queryParams: QueryParams;
}
declare type FakeRequest = FakeXMLHttpRequest & ExtraRequestData;
declare class NoopArray {
    length: number;
    push(..._items: any[]): number;
}
export default class Pretender {
    static parseURL: typeof parseURL;
    static Hosts: typeof Hosts;
    static Registry: typeof Registry;
    hosts: Hosts;
    handlers: ResponseHandler[];
    handledRequests: any[] | NoopArray;
    passthroughRequests: any[] | NoopArray;
    unhandledRequests: any[] | NoopArray;
    requestReferences: any[];
    forcePassthrough: boolean;
    disableUnhandled: boolean;
    ctx: {
        pretender?: Pretender;
    };
    running: boolean;
    private _nativeXMLHttpRequest;
    private _fetchProps;
    constructor();
    get: (this: Pretender, path: string, handler: ResponseHandler, async: boolean) => ResponseHandlerInstance;
    post: (this: Pretender, path: string, handler: ResponseHandler, async: boolean) => ResponseHandlerInstance;
    put: (this: Pretender, path: string, handler: ResponseHandler, async: boolean) => ResponseHandlerInstance;
    delete: (this: Pretender, path: string, handler: ResponseHandler, async: boolean) => ResponseHandlerInstance;
    patch: (this: Pretender, path: string, handler: ResponseHandler, async: boolean) => ResponseHandlerInstance;
    head: (this: Pretender, path: string, handler: ResponseHandler, async: boolean) => ResponseHandlerInstance;
    options: (this: Pretender, path: string, handler: ResponseHandler, async: boolean) => ResponseHandlerInstance;
    map(maps: (pretender: Pretender) => void): void;
    register(verb: string, url: string, handler: ResponseHandler, async: boolean): ResponseHandlerInstance;
    passthrough: {};
    checkPassthrough(request: FakeRequest): boolean;
    handleRequest(request: FakeRequest): void;
    handleResponse(request: FakeRequest, strategy: any, callback: Function): void;
    resolve(request: FakeRequest): void;
    requiresManualResolution(verb: string, path: string): boolean;
    prepareBody(body: any, _headers: any): any;
    prepareHeaders(headers: any): any;
    handledRequest(_verb: any, _path: any, _request: any): void;
    passthroughRequest(_verb: any, _path: any, _request: any): void;
    unhandledRequest(verb: any, path: any, _request: any): void;
    erroredRequest(verb: any, path: any, _request: any, error: any): void;
    shutdown(): void;
    private _handlerFor;
}
export {};

```

</details>

katehedgpeth:
@xg-wang Great suggestion. I've added types to the class methods, and exported the old `Server` type from `pretender.ts`. It looks like the build is failing now though, and I'm not sure how to fix it. Do I need to bump the version number?

src/interceptor.ts
```diff
@@ -2,7 +2,7 @@ import FakeXMLHttpRequest from 'fake-xml-http-request';
 import { createPassthrough } from './create-passthrough';
 
 export function interceptor(ctx) {
-  function FakeRequest() {
+  function FakeRequest(this: any) {
```
katehedgpeth, line 5: I really hate using `any`, but I wasn't totally sure what this should be. Is it `Pretender`?

package.json
```diff
@@ -42,6 +42,7 @@
     "release-it-lerna-changelog": "^2.3.0",
     "rollup": "^2.10.2",
     "rollup-plugin-typescript": "^1.0.0",
```
xg-wang, line 44: I think you need to remove this then rerun `yarn`.

> error Your lockfile needs to be updated

xg-wang (changes requested):

src/interceptor.ts
xg-wang: This should really be rewritten into class :) Seems the generated type is fine, maybe just leave it there for now

PR #353, opened 2022-06-27
Add handledRequests, passthroughRequests & unhandledRequests to ts def

### LINKED ISSUE #337: Server class in index.d.ts does not match Pretender class

abel-n:
Fix https://github.com/pretenderjs/pretender/issues/337

Moving our project to typescript I've bumped into an issue that's been there for a long time apparently: missing handled / unhandled / passthrough request information from the typescript definition.

I can see that the actual value might be an instance of `NoopArray`, but type-wise I think it should be safe to just stick to using an array of requests. Suggestions are welcome. :relaxed:

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -9,6 +9,10 @@ export type Config = SetupCallback | SetupConfig;
 export class Server {
   public passthrough: ResponseHandler;
 
+  public handledRequests: (FakeXMLHttpRequest & ExtraRequestData)[];
+  public passthroughRequests: (FakeXMLHttpRequest & ExtraRequestData)[];
+  public unhandledRequests: (FakeXMLHttpRequest & ExtraRequestData)[];
+
   constructor(config?: Config);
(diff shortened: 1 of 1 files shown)


tuzmusic:
Any chance this will be merged soon?

Issue #361, opened 2023-04-04
Why is the polyfill `fetch` library being used always?


bobziroll:
There's a conflict with React Router ≥ v6.4.5, where it always expects a response to have a `body` property. However, the `whatwg-fetch` library used by pretender is intercepting the `Response` created by React Router and appending a `_bodyInit` property instead of a `body` property. So React Router checks this Response with a duck-typing check, sees that it doesn't have a `body` property (which the W3 spec requires of a Response obj), and rejects it as not being a valid Response.

(In v6.4.4 they were checking it with `instanceof Request` which was passing that check.)

My question is, why is this polyfilled `fetch` being used instead of the native fetch when my browser has a native fetch? The native fetch doesn't have this problem.

mdbiscan:
Is this happening for you in Node?

bobziroll:
I haven't tried it in Node. I realize now I wasn't very descriptive of my scenario.

I'm building a React project w/ React Router and using Mirage JS as a fake API. Mirage JS is using Pretender, which is using `whatwg-fetch` as the polyfill for fetch. Because that polyfill doesn't include a `body` property, React Router isn't recognizing the Response as a legitimate response, since Response objects need to have a true `.body` property according to the spec.

There are some limitations to the polyfill and streaming responses, which is why they chose not to include a true `.body` on their Responses. However, from poking around the Pretender code base, it looks like that polyfill is being used whether the browser environment it's being used in is modern enough to have a `fetch` implementation built in.

I wrote [this little PR](https://github.com/pretenderjs/pretender/pull/362) that would check to see if the browser being used has `window.fetch` already.

I also might be way in over my head with this 😅 but I thought I'd give it a shot in the dark to see if that might help avoid using the polyfill if it isn't needed in that environment. But I'm very happy to be corrected if I'm not going about things the right way.

vincicat:
I also have similar issue with the fetch polyfill of pretender in node.js (v20.0) - `global.Headers` is now a read only property as fetch is officially supported for a while (see [link](https://blog.logrocket.com/fetch-api-node-js/)), the mocking always throw a Type Error now.

Please accept the PR #362

na-ji:
[...]

PR #362, opened 2023-04-04
Check if `window.fetch` exists before polyfilling it

bobziroll:
Sorry if this is presumptuous of me, I haven't done much by way of contributing code to OS before, so please let me know if I need to do something before submitting this.

I'm running into an [issue](https://github.com/pretenderjs/pretender/issues/361#issue-1653170925) where `whatwg-fetch` is hijacking the Response object that React Router checks when redirecting. React Router ≥v6.4.5 duck-type checks the Response object, checking specifically if there is a `body` property included. (Since it's [required by W3 Spec](https://fetch.spec.whatwg.org/#concept-response-body)).

I guess `whatwg-fetch` doesn't support the body property, but instead injects a `_bodyInit` property in its place. Something to do with ReadableStreams or whatnot, I'm not sure 😝

That said, I was curious why pretender is using the polyfill in the first place, since my browser is a modern, `fetch`-enabled browser.

Please let me know if there's another, better way I could contribute to this, or help me better understand why a PR like this doesn't make sense. Thanks in advance! 🙏

### DIFF
--- a/src/pretender.ts
+++ b/src/pretender.ts
@@ -120,13 +120,15 @@ export default class Pretender {
     (<any>self).XMLHttpRequest = interceptor(ctx);
 
     // polyfill fetch when xhr is ready
-    this._fetchProps = FakeFetch
-      ? ['fetch', 'Headers', 'Request', 'Response']
-      : [];
-    this._fetchProps.forEach((name) => {
-      (<any>this)['_native' + name] = self[name];
-      self[name] = FakeFetch[name];
-    }, this);
+    if (!self.fetch) {
+      this._fetchProps = FakeFetch
+        ? ["fetch", "Headers", "Request", "Response"]
+        : [];
+      this._fetchProps.forEach((name) => {
+        (<any>this)["_native" + name] = self[name];
+        self[name] = FakeFetch[name];
+      }, this);
+    }
 
     // 'start' the server
     this.running = true;


vincicat:
Please accept this PR as it will solve the Pretender Error in node.js 18+

vincicat:
@bobziroll Thank you for your help.

I got error in other line that iterating `this._fetchProps` (e.g. dist/pretender.js:1809:27), may be `this._fetchProps` need to be an empty array by default to ensure no more code will be broken...

fekete965:
> @bobziroll Thank you for your help.
> 
> I got error in other line that iterating `this._fetchProps` (e.g. dist/pretender.js:1809:27), may be `this._fetchProps` need to be an empty array by default to ensure no more code will be broken...

Yes indeed, at the top of the constructor we need to initialize it to an empty array.

I would also propose to validate the incoming verbs:
```ts
const Verbs: Record<string, Verb> = {
  GET: 'GET',
  POST: 'POST',
  PUT: 'PUT',
  DELETE: 'DELETE',
  PATCH: 'PATCH',
  HEAD: 'HEAD',
  OPTIONS: 'OPTIONS',
} as const;

const validatedVerb = (str: string): Verb => {
  str = str.toUpperCase();

  if (!Verbs[str]) {
    const validVerbs = Object.values(Verbs).join(', ');
    throw new Error(
      `${str} is not a valid verb. Expected value one of [${validVerbs}]`
    );
  }

  return Verbs[str];
};
```

I also believe the `map` on the `Pretender` instance itself is wrongly typed:
```ts
map(fn: (pretender: Pretender) => void) {
  n.call(this);
}
```
I believe this should be:
```ts
map(fn: (this: Pretender) => void) {
  fn.call(this);
}
```
This way we tie what context the function can be called with.

A better API would be:
```ts
map(fn: (this: Pretender) => void): Pretender {
  fn.call(this);
  
  return this;
}
```
This way we could nicely chain our operations together:
```ts
const myPretender = new Pretender()
  .map(authenticationRoutes);
  .map(songsRoutes);
```

Now I can also see another Type error in the following line:
```ts
requiresManualResolution(verbStr: string, path: string) {
  let verb = validatedVerb(verbStr);
  let handler = this._handlerFor(verb, path, {}); <-- object is not a valid argument
  // etc...
}
```
`this._handlerFor` accepts a `FakeRequest` but an empty object is not valid.
I am not sure that this is intended or that we should have a factory that creates fake requests for us.

Sorry @bobziroll if I spammed your PR here. 😅

PR #355, opened 2022-06-27
Fix passthrough issue for libraries using loadend instead of onreadystatechange (Axios)

DrewCerny:
Issue #354 Fixed by adding loadend to the list of events watched by the passthrough

PR #363, opened 2024-01-29
fix: passthrough for binary files

### LINKED ISSUE #72: Binary files are converted to strings

### LINKED ISSUE #305: Passthrough for binary files not working
I just spent many hours trying to figure out why a binary file sent via an API was always corrupted, until I found out it was because of Pretender passthrough somehow bungling it up.

I searched and found this: https://github.com/pretenderjs/pretender/pull/148, but as far as I see this cannot be set when using `fetch()`.

### LINKED ISSUE #1915: Mirage-overridden fetch cannot download an image binary blob
# Mirage has a bug when using the [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) API to download an image [`Blob`](https://developer.mozilla.org/en-US/docs/Web/API/Blob).

Here is a repo that reproduces the problem: https://github.com/skarger/mirage-blob-fetch.
It's a simple Ember app that can be run with `yarn start`.

The Mirage-overridden `fetch` function seems to return a *text* version of the image's bytes, instead of the desired `Blob`. Same problem if trying to access the image as an `ArrayBuffer`. 

The fetched data is corrupt, and also larger. I believe it's larger because the text representation of the image's bytes is UTF-8 encoded, and therefore each of its "characters" may expand in width from 1 byte to up to 4 bytes.

There are a few players in the mix:

* [`Mirage`](https://github.com/miragejs/ember-cli-mirage)
* [`Pretender`](https://github.com/pretenderjs/pretender), which Mirage uses to mock the `fetch` function.
* [`whatwg-fetch`](https://github.com/github/fetch), a polyfill for `window.fetch`, which [Pretender uses to
overwrite the browser's native `fetch` function](https://github.com/pretenderjs/pretender/blob/46145a51c6b6900b24517793929c8c2c1fa4ba20/src/index.ts#L64).

* I believe the bug only happens when using Mirage.
* I've confirmed it does *not* happen when only using `whatwg-fetch`.

The example app's index route provides a demo that reproduces the problem.

Clicking the buttons will show the returned blob type and byte sizes when using
the native `fetch` function and when using the Mirage-overridden `fetch`.

You can see they are different. Clicking these buttons prompts console logging as well.

Demo app screenshot:
mirage-blob-fetch

In order to make this demo, I used an Ember application initializer to capture
the native `fetch` function as `window.nativeFetch`. Otherwise I would have no way to access
native `fetch` because Mirage / Pretender overrides it.

Techn1x:
Fixes https://github.com/pretenderjs/pretender/issues/305
Fixes https://github.com/pretenderjs/pretender/issues/72
Fixes https://github.com/miragejs/ember-cli-mirage/issues/1915

I don't actually fully understand the change being made here, I've just converted the fix from this comment into a PR https://github.com/pretenderjs/pretender/issues/305#issuecomment-685815226
It works great for me. Full credit to @ankology
* I left off the tests from that comment, not sure exactly how to integrate them as tests on this repo

Happy to adjust this PR to update with any suggestions. Would love for it to be merged, but even if it doesn't it at least provides an easier way for people to see the code diff & write their own pnpm patches or point their repos at this branch.

---

Unrelated, but I would implore this fix PR be merged as well https://github.com/pretenderjs/pretender/pull/355
It's another passthrough bug that I've had to write a pnpm patch for in my projects.

### DIFF
--- a/src/create-passthrough.ts
+++ b/src/create-passthrough.ts
@@ -25,7 +25,7 @@ export function createPassthrough(fakeXHR, nativeXMLHttpRequest) {
     fakeXHR.password
   );
 
-  if (fakeXHR.responseType === 'arraybuffer') {
+  if (fakeXHR.responseType === 'blob' || fakeXHR.responseType === 'arraybuffer') {
     lifecycleProps = ['readyState', 'response', 'status', 'statusText'];
     xhr.responseType = fakeXHR.responseType;
   }
@@ -37,7 +37,7 @@ export function createPassthrough(fakeXHR, nativeXMLHttpRequest) {
 
   // add progress event for async calls
   // avoid using progress events for sync calls, they will hang https://bugs.webkit.org/show_bug.cgi?id=40996.
-  if (fakeXHR.async && fakeXHR.responseType !== 'arraybuffer') {
+  if (fakeXHR.async) {
     evts.push('progress');
     uploadEvents.push('progress');
   }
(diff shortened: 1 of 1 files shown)


Techn1x:
@xg-wang as requested https://github.com/pretenderjs/pretender/issues/305#issuecomment-826106443

Let me know if there's anything I can do to get this merged.

Techn1x:
MSW is the way forward

